PR #3, opened 2024-10-19
Fix: Update pip install command for zsh compatibility

picsoung:
Original `pip install` command was giving errors on zsh `zsh: no matches found: pipecat-ai[daily,openai]`

Adding quotes help prevent that and make the installation smooth 👌


DominicStewart (approved):
Yes, I remember running into this when I was first learning Pipecat! Good catch!

PR #7, opened 2024-10-27
Initial docs creation

markbackman:
Lots of docs updates:
- Getting started sections
- Core concepts
- API Reference (generated from services and types files)
- Cleaned up the navigation


core-concepts/overview.mdx
```diff
@@ -0,0 +1,127 @@
...
+5. Audio playback starts as soon as the first sentence is ready
+6. LLM context is aggregated and updated continuously and in real-time
+
+This streaming approach creates natural, responsive interactions.
+
+## Architecture Overview
+
+Here's how Pipecat organizes these processes:
+
+<Frame>![Platform architecture](/images/architecture-2.png)</Frame>
```
vipyne, line 29: re: this image- should we rename "AudioFrame" and "LLMContextFrame" as those are no longer frame types?

markbackman: I'm using InputAudioRawFrame for audio and I updated `LLMContextFrame` to `OpenAILLMContextFrame`, which is more accurate.

vipyne (approved):
this is  a w e s o m e .  thank you @markbackman !
I'm going to Approve because I think it is better than what we currently have.  it's pretty big and I'm sure there are a few typos, but we can make follow up PRs for those fixes.  (and thanks in advance to folks who catch things and make PRs!)

PR #9, opened 2024-10-29
docs: small nits and tweaks

vipyne:
mostly nits. The most important change imo is the `task = PipelineTask(pipeline)` addition to quickstart.


markbackman (approved):
Thanks for these changes! They're great. 🙇

PR #52, opened 2024-12-11
Android updates for 0.3.0

marcus-daily:


client/reference/android/api-reference.mdx
```diff
@@ -1,4 +1,4 @@
 ---
 title: "API Reference"
-url: "https://docs-android.rtvi.ai/rtvi-client-android/index.html"
```
marcus-daily, line 3: After the rename these directories will be different, so for now just point to the root.

markbackman (approved):
Thank you!

PR #78, opened 2024-12-23
Fixing iOS Docs

filipi87:
Fixing iOS Docs to match the current API.


markbackman:
Thanks for catching this!

markbackman (approved):

PR #116, opened 2025-02-04
Update Flows state docs to include results data

markbackman:
Docs for: https://github.com/pipecat-ai/pipecat-flows/pull/92


markbackman:
Changed implementation. Closing.

PR #127, opened 2025-02-10
Update pipecat flows pre and post action signatures

kompfner:
This change is needed after [this change](https://github.com/pipecat-ai/pipecat-flows/pull/93) lands in `pipecat-flows` and is published.


markbackman (approved):
LGTM!

PR #140, opened 2025-02-26
Add Flows start_callback docs

markbackman:
Docs for: https://github.com/pipecat-ai/pipecat-flows/pull/101

### DIFF
--- a/guides/features/pipecat-flows.mdx
+++ b/guides/features/pipecat-flows.mdx
@@ -289,14 +289,13 @@ Functions that create transitions between nodes. When called, they:
 Start callbacks execute before a function is called, allowing you to:
 
 - Provide immediate feedback while long-running operations complete
-- Acknowledge user requests before processing
-- Add wait messages for API calls or database operations
+- Execute logic upon a function call's initiation
 
 ```python
 # Define a start callback
-async def before_check_weather(function_name: str, llm: Any, context: Any) -> None:
+async def before_check_weather(function_name: str, flow_manager: FlowManager) -> None:
     """Tell the user we're checking the weather."""
-    await llm.push_frame(TTSSpeakFrame("Let me check the weather for you..."))
+    await flow_manager.llm.push_frame(TTSSpeakFrame("Let me check the weather for you..."))
 
 # Use in function configuration
 {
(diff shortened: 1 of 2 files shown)


guides/features/pipecat-flows.mdx
```diff
@@ -274,16 +278,45 @@ Functions that create transitions between nodes. When called, they:
...
         "transition_to": "target_node" # Required: Specify next state
     }
 }
 ```
 
+#### Start Callbacks
+
+Start callbacks execute before a function is called, allowing you to:
+
+- Provide immediate feedback while long-running operations complete
```
kompfner, line 291: These 3 bullet points sound kind of same-ish to me, like they're just paraphrasing (like I just did with this feedback whoops haha)

guides/features/pipecat-flows.mdx
```diff
@@ -274,16 +278,45 @@ Functions that create transitions between nodes. When called, they:
...
+    }
+}
+```
+
 A function's behavior is determined by its properties:
 
 - `handler` only: Process data, stay in current state
 - `transition_to` (static flow) or `transition_callback` (dynamic flow) only: Pure transition to next state
-- Both: Process data, then transition
+- `start_callback` only or combined with others: Provide immediate feedback before execution
```
kompfner, line 318: Are we sure you can provide `start_callback` by itself? Or does it need to piggyback on a provided `handler`? It kind of doesn't make much conceptual sense on its own...

guides/features/pipecat-flows.mdx
kompfner: Same feedback applies to where these bullets appear again

guides/features/pipecat-flows.mdx
markbackman: I suppose it's possible but unlikely.

When there's no handler, it's likely an edge function and just transitioning, which shouldn't require a start_callback. But, it's possible 🤷.

markbackman: I'm just going to drop the only/combined part.

guides/features/pipecat-flows.mdx
markbackman: Improved.

filipi87 (approved):
LGTM!

kompfner (approved):

PR #147, opened 2025-03-14
Improve docs for DailyTransport

mattieruth:


client/js/transports/daily.mdx
```diff
@@ -2,22 +2,140 @@
...
+  const authBundle = await fetch(connectUrl, {...});
+  expect(typeof authBundle.room_url).toBe('string');
+  if (authBundle.token) {
+    expect(typeof authBundle.token).toBe('string');
+  }
+  ```
+
+- `preAuth()`
+
+  This is the one method meant to be called directly, which is used to allow you to gather information about the Daily room prior to connecting. As a Daily-specific action, it is not exposed through the RTVI client. This method must be called priort to `connect()` and use the same room_url and token (optional) as what will be returned by your connection endpoint.
```
markbackman, line 85:
```suggestion
  This is the one method meant to be called directly, which is used to allow you to gather information about the Daily room prior to connecting. As a Daily-specific action, it is not exposed through the RTVI client. This method must be called prior to `connect()` and use the same room_url and token (optional) as what will be returned by your connection endpoint.
```

client/js/transports/daily.mdx
```diff
@@ -2,22 +2,140 @@
...
+| 'error'
+```
+
+### Methods
+
+Most methods have an equivalent in the RTVI client. The transport is designed to be used with the RTVI client, so in most cases you should use the RTVI client methods to interact with the transport, rather than calling these methods directly. This is especially true for `initialize()`, `connect()`, and `disconnect()`.
+
+- `connect()`
+
+  This method is called by the RTVI client when it is time to connect to the Daily room. It should not be called directly. However, it does have some expectations. The RTVI client will pass the response from your connection endpoint to this method and expects the response to include a `room_url` and an optional `token`. See the [baseUrl](https://docs.pipecat.ai/client/js/api-reference/client-constructor#param-base-url) and [connect endpoint](https://docs.pipecat.ai/client/js/api-reference/client-constructor#param-endpoints) documentation for more information.
```
markbackman, line 73: Maybe emphasize this?
```suggestion
  This method is called by the RTVI client when it is time to connect to the Daily room. **It should not be called directly.** However, it does have some expectations. The RTVI client will pass the response from your connection endpoint to this method and expects the response to include a `room_url` and an optional `token`. See the [baseUrl](https://docs.pipecat.ai/client/js/api-reference/client-constructor#param-base-url) and [connect endpoint](https://docs.pipecat.ai/client/js/api-reference/client-constructor#param-endpoints) documentation for more information.
```

markbackman: Is it worth the detail on how this works if they don't need to use it and it happens automatically?

I do think the return information is useful, in general, so maybe that can be added somewhere either mentioned that this is the response from `connect()` or elsewhere on this page.

client/js/transports/daily.mdx
```diff
@@ -2,22 +2,140 @@
...
+
+## Events
+
+The transport implements the various [RTVI event handlers](https://docs.pipecat.ai/client/js/api-reference/callbacks). These events will not fire until the transport has been attached to an RTVIClient.
+
+## Advanced
+
+### Accessing the Daily Call
+
+For advanced use cases, where you may need to work with the Daily call object directly, you can access it via the `dailyCallClient` property.
```
markbackman, line 103:
```suggestion
For advanced use cases, where you may need to work with the Daily call client directly, you can access it via the `dailyCallClient` property.
```

client/js/transports/daily.mdx
```diff
@@ -2,22 +2,140 @@
...
+
+```javascript
+const dailyCall = transport.dailyCallClient;
+```
+
+<Note>
+The Daily client returned is safe-guarded to not allow you to call functions which affect the call's lifecycle and will redirect you to use either a Transport method or the RTVI client to perform the equivalent action.
```
markbackman, line 110:
```suggestion
The Daily call client returned is safe-guarded to not allow you to call functions which affect the call's lifecycle and will redirect you to use either a Transport method or the RTVI client to perform the equivalent action.
```

markbackman (approved):
Just a few simple comments. Otherwise, LGTM.

PR #165, opened 2025-03-20
update docs for client-js opeanai webrtc transport

mattieruth:


client/js/transports/openai-webrtc.mdx
```diff
@@ -37,30 +37,60 @@ let RTVIConfig: RTVIClientOptions = {
...
-      | "sage"
-      | "shimmer"
-      | "verse";
-    input_audio_transcription?: {
-      model: "whisper-1";
-    };
-    temperature?: number;
-    max_tokens?: number | "inf";
-  };
+export type OpenAIFunctionTool = {
```
markbackman, line 40: Is this the right format for this type of documentation? I'm not sure of the context. In other places we use ParamFields to document.

jptaylor (approved):

client/js/transports/openai-webrtc.mdx
mattieruth: I don't think that makes sense here. I don't want to be in the business of re-writing OpenAI's docs about all their configs. This is a nice sort-of self explanatory way to show what's possible (and is copy-pasted from what we've implemented). I added a quick sentence at the top to point people to their docs so they can get more detail :)

client/js/transports/openai-webrtc.mdx
```diff
@@ -36,31 +36,63 @@ let RTVIConfig: RTVIClientOptions = {
 
 ### Configuration Options
 
+Below is the transport's type definition for the OpenAI Session configuration. See the [OpenAI Realtime API documentation](https://platform.openai.com/docs/api-reference/realtime-client-events/session/update) for more detauks on the eacg if the options and their defaults.
```
markbackman, line 39: @mattieruth Sorry, I should have caught this before. There are two typos in this paragraph:

```
for more detauks on the eacg if the options and their defaults.
```

mattieruth: no you shouldn't have. i pushed this after your review 🤦‍♀️ -- fix added to #169

PR #270, opened 2025-06-24
add watchdog timers

aconchillo:


markbackman (approved):
Nice clean up of the nav!

Looks good. I pushed a change with some typo fixes. Looks great.

PR #281, opened 2025-07-10
Update Flows reference docs and guide

markbackman:
Rewrite of both docs pages:
- we now have auto-generated reference docs, so we can slim down the reference docs
- we have a better idea of how to build with flows + we can shift the focus to dynamic flows

### DIFF
--- a/guides/features/pipecat-flows.mdx
+++ b/guides/features/pipecat-flows.mdx
@@ -189,7 +189,7 @@ Function handlers must return a tuple containing:
 - **Result**: Data provided to the LLM for context in subsequent completions
 - **Next Node**: The `NodeConfig` for Flows to transition to next
 
-Some handlers may not want to transition conversational state, in which case you can return `None` for the next node.
+Some handlers may not want to transition conversational state, in which case you can return `None` for the next node. Other handlers may _only_ want to transition conversational state without doing other work, in which case you can return `None` for the result.
 
 #### Direct Functions
 


guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
-    title="Insurance Policy (Dynamic)"
-    icon="shield"
-    href="https://github.com/pipecat-ai/pipecat-flows/blob/main/examples/dynamic/insurance_gemini.py"
+    title="Restaurant Reservation"
+    icon="utensils"
+    href="https://github.com/pipecat-ai/pipecat-flows/blob/main/examples/dynamic/restaurant_reservation.py"
   >
-    A dynamic flow using Google Gemini that adapts policy recommendations based
-    on user responses. Demonstrates runtime node creation and conditional paths.
+    A simple, practical dynamic flow example
```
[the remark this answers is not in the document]
kompfner: If we're de-emphasizing static v dynamic flows (with the latter being the "default" way of doing things), do we need to mention "dynamic" here?

guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
 
-## Designing Conversation Flows
+- Receive audio from a user
+- Transcribe the user's input
+- Run an LLM completion
+- Convert the LLM response to audio
+- Send audio back to the user
 
-Functions in Pipecat Flows serve two key purposes:
+**Pipecat Flows** complements Pipecat's core functionality by managing the conversation's context and tools. This is separate from the core pipeline, allowing you to separate the conversation logic from the core pipeline mechanics.
```
kompfner, line 39: IMO, "managing the conversation's context and tools" kind "undersells" what the framework does. It doesn't say to what end it does those things.

Taking a stab at summarizing:

```suggestion
**Pipecat Flows** complements Pipecat's core functionality by providing a layer of structure to a conversation, managing context and tools as the conversation moves from one state to another. This is separate from the core pipeline, allowing you to separate conversation logic from core pipeline mechanics.
```

guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
-- Advancing the conversation to the next node
+## When to Use Pipecat Flows?
 
-Each function can do one or both.
+Pipecat Flows is best suited for use cases where:
 
-LLMs decide when to run each function, via their function calling (or tool calling) mechanism.
+- You require precise control over how a conversation progresses
+- Your bot needs to handle a complex task that can be decomposed into discrete steps
+- You're trying to improve the accuracy of your LLM's response or tool use
```
kompfner, line 47: Not sure I grok this 3rd point. Can you explain?

kompfner: Ah, you're referring to cases where you might otherwise have a huge prompt and many tools registered at once. I think I understand. The follow sentence makes it clearer.

guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
-If the function advances the conversation to the next node, it should return a non-`None` value for the second element of the tuple. This value can be either:
-
-- A `NodeConfig` defining the next node (for dynamic flows)
-- A string identifying the next node (for static flows)
-
-### Example Function
-
-```python
-from pipecat_flows import FlowArgs, FlowManager, FlowResult, NodeConfig
+<Note>
```
kompfner, line 58: 👍

guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
 
-async def check_availability(args: FlowArgs, flow_manager: FlowManager) -> tuple[FlowResult, NodeConfig]:
-    # Read arguments
-    date = args["date"]
-    time = args["time"]
+## Technical Overview
 
-    # Read previously-stored data
-    party_size = flow_manager.state.get("party_size")
+Pipecat Flows represents a conversation as a graph, where each step of the conversation is represented by a node. Nodes are typed as a `NodeConfig`, which may contain the following properties:
```
kompfner, line 67: nit

```suggestion
Pipecat Flows represents a conversation as a graph, where each step of the conversation is represented by a node. Nodes are of type `NodeConfig`, and may contain the following properties:
```

guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
 
-    # Use flow_manager for immediate user feedback
-    await flow_manager.task.queue_frame(TTSSpeakFrame("Checking our reservation system..."))
+- `name`: The name of the node; used as a reference to transition to the node
+- `role_messages`: A list of message `dicts` defining the bot's role/personality
```
kompfner, line 70: Should we mention this should only be used on the first node?

guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
-3. Do both
+### Functions
 
-This leads to two conceptual types of functions:
+Functions in Pipecat Flows serve two key purposes:
 
-- **Node functions**, which only process data.
-- **Edge functions**, which also (or only) transition to the next node.
+1. **Process data** by interfacing with external systems and APIs to read or write information
+2. **Navigate the conversation** by transitioning between nodes in your flow
```
kompfner, line 133: nit


```suggestion
2. **Progress the conversation** by transitioning between nodes in your flow
```

guides/features/pipecat-flows.mdx
```diff
@@ -466,517 +292,273 @@ NodeConfig(
...
 
-### Best Practices
+### Function Calls
 
-- Store shared data in flow_manager.state
-- Create separate functions for node creation
+Pipecat Flows supports three ways to define function calls:
 
-# Flow Editor
+#### FlowsFunctionSchema (Recommended)
```
kompfner, line 419: As I've been reading this doc and looking at our examples, I'm wondering if/when we should flip to recommending direct functions as the main recommended way of doing things, with function schemas as an alternative if you happen to need more fine-grained control (mainly over the `properties` object). Using functions directly really does cut down boilerplate and opportunities for making mistakes...

guides/features/pipecat-flows.mdx
```diff
@@ -466,517 +292,273 @@ NodeConfig(
...
-    ```
-  - Node Functions (Orange): Perform operations
-    ```python
-    {
-        "name": "process_data",
-        "handler": process_data_handler,
-        "description": "Process user data"
-    }
-    ```
+Uses the provider's native format (OpenAI, Anthropic, etc.). While supported, we recommend using FlowsFunctionSchema for better portability.
```
kompfner, line 469: 👍 I like how here (and above, with the static flows example), we're cutting down on the time we spend on non-recommended ways of doing things.

guides/features/pipecat-flows.mdx
```diff
@@ -466,517 +292,273 @@ NodeConfig(
...
 
-## Naming Conventions
+### Actions
```
kompfner, line 471: Wait...weren't we talking about actions earlier in this doc? Why did we circle back to it? (A little lost on the structure)

guides/features/pipecat-flows.mdx
```diff
@@ -466,517 +292,273 @@ NodeConfig(
...
+"pre_actions": [
+    {
+        "type": "tts_say",
+        "text": "Please hold while I process your request..."
+    }
+]
 ```
 
-When using the Flow Editor, function handlers can be specified using the `__function__:` token:
+#### Custom Actions
```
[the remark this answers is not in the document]
kompfner: Also: I had recently renamed "built-in" to "pre-canned" in the reference docs to distinguish between 3 types:
* pre-canned
* function
* custom (to be used rarely)

server/frameworks/flows/pipecat-flows.mdx
```diff
@@ -1,1125 +1,135 @@
...
+- **ActionError** - Action execution failures
 
-</CodeBlock>
+## Additional Notes
 
-</ResponseField>
+- **State Management**: Use `flow_manager.state` dictionary for persistent conversation data
+- **Automatic Function Call Registration and Validation**: All functions are automatically registered and validated at run-time
+- **Provider Compatibility**: Format differences handled automatically via adapter system
+- **Deprecation**: Some legacy patterns (transition_to, transition_callback) are deprecated in favor of consolidated function handlers
```
kompfner, line 135:
```suggestion
- **Deprecation**: Some legacy patterns (`transition_to`, `transition_callback`) are deprecated in favor of consolidated function handlers
```

guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
-        "parameters": {
-            "type": "object",
-            "properties": {
-                "size": {"type": "string", "enum": ["small", "medium", "large"]}
-            }
-        },
-    }
-}
-```
+Some handlers may not need to return a next node, in which case you can return `None`. Also, some handlers may not need to return a result, in which case you can return `None` for the result as well.
```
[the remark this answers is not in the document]
markbackman: Maybe we should encourage using None only for the case where you want to do work in the node and not transfer. I think you pretty much always want to return a result. I'll make that update.

guides/features/pipecat-flows.mdx
```diff
@@ -1,441 +1,267 @@
...
 
-Note that when the node is configured with `respond_immediately: False`, the `pre_actions` still run when entering the node, which may be well before LLM inference, depending on how long the user takes to speak first.
+#### Custom Actions
 
-<Tip>
-  Avoid mixing `tts_say` actions with chat completions as this may result in a
-  conversation flow that feels unnatural. `tts_say` are best used as filler
-  words when the LLM will take time to generate an completion.
-</Tip>
+You can define your own actions to handle specific business logic or integrations. Custom actions give you complete flexibility to execute any functionality your application needs.
```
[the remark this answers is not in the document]
markbackman: Good point.

guides/features/pipecat-flows.mdx
```diff
@@ -466,517 +292,273 @@ NodeConfig(
...
-```
-
-## Dynamic Flows
+<Note>For Static Flows, `initialize()` does not take any args.</Note>
 
-Dynamic flows create and modify conversation paths at runtime based on data or business logic.
+## State Management
 
-### Example Implementation
+Pipecat Flows includes built-in state management through the `flow_manager.state` dictionary. This persistent storage lets you share data across nodes throughout the entire conversation.
```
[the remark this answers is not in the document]
markbackman: Yeah! Cross-Node State is a great concept. The idea is to make it easier to have state to pass around, avoiding the need for globals.

guides/features/pipecat-flows.mdx
markbackman: I set this up as:
- Technical overview (all sections)
- Usage (all sections)

Though, the actions examples are trivial, so I'll integrate them into the technical overview section.

guides/features/pipecat-flows.mdx
kompfner: Yes! Reads great now

guides/features/pipecat-flows.mdx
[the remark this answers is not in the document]
kompfner: > I think you pretty much always want to return a result

I don't think that's true. We would want to return only a next node and no result in all the cases where we'd previously use a `transition_*` by itself without a `handler`. Fixing.

guides/features/pipecat-flows.mdx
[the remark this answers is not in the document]
kompfner: > Maybe we even deprecate custom actions in favor of using functions. WDYT? It should be functionally equivalent.

That might be a good idea. Could minimize confusion, for sure. 

They're technically not _completely_ functionally equivalent, as a custom action as the first action in the `post_actions` list would run _immediately_ after the LLM completion whereas a function action in the same position would run after the bot finished speaking.

Regardless, let's revisit this question after this PR.

PR #286, opened 2025-07-23
Fixing the docs for the TavusVideoService and TavusTransport.

filipi87:


markbackman (approved):
LGTM! Thanks 🙌

filipi87:
@markbackman, I have just added the docs for the `TavusTransport` as well, that was missing. 

Let me know what you think.

server/services/transport/tavus.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: Tavus Transport
+order: 20
+---
+
+import { Preview } from 'src/components/Preview'
```
markbackman, line 6: These two imports shouldn't be needed.

server/services/transport/tavus.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: Tavus Transport
+order: 20
```
markbackman, line 3: What does `order` do? (I'm always learning something new about Mintlify!)

server/services/transport/tavus.mdx
```diff
@@ -0,0 +1,96 @@
...
+---
+
+import { Preview } from 'src/components/Preview'
+import { CodeBlock } from 'src/components/Code'
+
+Connects your Pipecat app to a Tavus conversation, allowing the bot to join the same virtual room as the Tavus avatar and participants.
```
markbackman, line 9: Let's move this to the title section, like this:
```
---
title: Tavus Transport
description: "Connects your Pipecat app to a Tavus conversation, allowing the bot to join the same virtual room as the Tavus avatar and participants."
order: 20
---
```

server/services/transport/tavus.mdx
filipi87: Yeah, but in this case, just my mistake. It shouldn't be needed here.  🤕

PR #312, opened 2025-08-01
Add documentation for device errors and update transport constructors…

mattieruth:
… that now take an optional media manager

Draft because this should not go in until these PRs (https://github.com/pipecat-ai/pipecat-client-web/pull/134 & https://github.com/pipecat-ai/pipecat-client-web-transports/pull/51) have been released


markbackman (approved):
Nice!

PR #338, opened 2025-08-29
Update client SDK 1.0 transition note

markbackman:


mattieruth (approved):
i approve but we might wait to modify until all the docs are updated and migration guides are provided for all platforms?

PR #727, opened 2026-04-15
Standardize docs install commands on uv

jamsea:
## Summary

- Replaces `pip install` with `uv add` across all docs install snippets (`uv pip install` for `-r requirements.txt`, `uv tool install` for CLI installs).
- Adds a new canonical [Installation](/pipecat/get-started/installation) page under Get Started that explains why uv is the default and links to PyPI as the source of truth for current package versions.
- Bumps Python requirement from 3.10 to 3.11 in the quickstart and CLI overview to match `pipecat-ai` 1.0 and `pipecat-ai-cli` 1.0 on PyPI.

## Why

Kapa coverage gap flagged that users couldn't find `pip install` instructions for Pipecat or know which version the docs target. Rather than maintain a static version table that goes stale (the gap that triggered this), the new Installation page points to PyPI for live versions and standardizes the tool used in every example.

The "Why uv" callout on the Installation page explains the rationale (faster, manages Python versions, lockfile by default, and matches what `pipecat init` scaffolds).

## Scope

- 137 files touched (1 new, 136 modified).
- Intentionally **not** changed:
  - `agent-images.mdx` Dockerfile-using-pip tab (paired alternative).
  - Build log JSON sample showing actual pip output.
  - Prose mentions like "Failed `pip install` commands" in build log docs.

## Test plan

- [x] `npx mint broken-links` passes
- [x] Prettier formatted
- [ ] Spot-check the new Installation page renders correctly in `mint dev`
- [ ] Verify the supported-services.mdx tables still align after `uv add` substitution


markbackman (approved):
LGTM! Thank you.

PR #744, opened 2026-04-17
Rework client sdk reference

mattieruth:
This moves 99% of existing client docs into the API Reference tab and introduces new higher level guides for the client sdks under that tab. When it makes sense, the guides have "Views" for toggling between sdks to get sdk-specific versions. Currently, the pages have views for Javascript, React, and React Native. @filipi87 -- would you mind reviewing the ReactNative-specific bits?


markbackman (approved):
Looks great!

One last thing is to add link redirects for the pages you've moved.

PR #747, opened 2026-04-17
Client guides ios

mattieruth:


filipi87 (approved):
Hi @mattieruth, thank you for doing this.

I have included in this follow up [PR](https://github.com/pipecat-ai/docs/pull/771) the changes I believe should be merged into this PR for the iOS docs.

And with this, I think we are good to go. 🚀

PR #749, opened 2026-04-17
Client guides android

mattieruth:


mattieruth:
i meant to make this PR against the ios branch so that the android-specific changes were clearer. replaced this PR with #754

PR #750, opened 2026-04-20
docs: update subagents docs for PR #16

markbackman:
Automated documentation update for [pipecat-subagents PR #16](https://github.com/pipecat-ai/pipecat-subagents/pull/16).

### api-reference/pipecat-subagents/decorators.mdx
- Removed `parallel` parameter from `@task` decorator Parameters section
- Removed code example showing `@task(name="research", parallel=True)`
- Added note clarifying that each task request runs in its own asyncio task and multiple tasks can be in flight simultaneously

### subagents/learn/task-coordination.mdx
- Removed section explaining `parallel=True` parameter (lines 91-98)
- Updated note in "Overriding on_task_request" section to remove reference to `@task(parallel=True)`

## Summary

PR #16 removed the `parallel` parameter from the `@task` decorator because all task handlers now always run in their own asyncio task by default. This was a breaking change that simplifies the API while ensuring the bus message loop is never blocked.

## Gaps identified

None. All relevant documentation has been updated.

### DIFF
--- a/api-reference/pipecat-subagents/decorators.mdx
+++ b/api-reference/pipecat-subagents/decorators.mdx
@@ -81,12 +81,6 @@ async def on_task_request(self, message):
 async def on_research(self, message):
     result = await do_research(message.payload)
     await self.send_task_response(result)
-
-# Parallel handler (each request runs concurrently)
-@task(name="research", parallel=True)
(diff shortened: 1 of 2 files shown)


markbackman:
Already implemented.

PR #753, opened 2026-04-20
Document maxSessionDuration configuration

mattshep:
Add reference docs for the new per-service max session duration field across the CLI (--max-session-duration flag + TOML config), REST API (agent create/update request + response schemas), and the Fundamentals/Starting Sessions guide. Also add a bullet in Capacity Planning calling out short caps as a cost safety net.

Use direct language about the enforcement behavior: sessions are forcibly closed when the limit is reached, not gracefully terminated — the bot sees a connection error on next I/O with no prior signal.


markbackman (approved):
LGTM!

PR #791, opened 2026-05-05
Fix --base64encode flag drift in image pull secret docs

jamsea:
## Summary

The CLI flag for `pipecat cloud secrets image-pull-secret` is `--encode` (default `true`), not `--base64encode`. Three pages referenced the wrong flag name and framed encoding as opt-in when it's actually the default.

Updates the wording on all three to match the CLI reference: encoding happens by default; pass `--encode=false` to opt out.

## Files changed

- `pipecat-cloud/fundamentals/secrets.mdx`
- `pipecat-cloud/guides/container-registries/docker-hub.mdx`
- `pipecat-cloud/security/security-and-compliance.mdx` (two occurrences)

## Test plan

- [x] `grep -rn base64encode docs` returns 0 hits (verified locally)
- [x] Mintlify build of the three pages renders the corrected wording

### DIFF
--- a/pipecat-cloud/fundamentals/secrets.mdx
+++ b/pipecat-cloud/fundamentals/secrets.mdx
@@ -146,7 +146,7 @@ pipecat cloud secrets image-pull-secret <secret-name> <registry-url>
 ```
 
 Running this command will prompt you for account credentials.
-You can optionally encode your credentials in base64 with the `--base64encode` flag.
+Credentials are base64-encoded by default; pass `--encode=false` to skip encoding if your credentials are already encoded.
 
 Like regular secret sets, image pull secrets are region-specific. Specify the region with the `--region` flag:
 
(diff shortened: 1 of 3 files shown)


pipecat-cloud/fundamentals/secrets.mdx
```diff
@@ -146,7 +146,7 @@ pipecat cloud secrets image-pull-secret <secret-name> <registry-url>
 ```
 
 Running this command will prompt you for account credentials.
-You can optionally encode your credentials in base64 with the `--base64encode` flag.
+Credentials are base64-encoded by default; pass `--encode=false` to store them without base64 encoding.
```
markbackman, line 149: --encode=false isn't an option.

markbackman (commented):
I'm going to close this out and open a separate PR with a slightly different solution:
https://github.com/pipecat-ai/docs/pull/797

PR #826, opened 2026-05-15
docs: update for pipecat PR #4501

markbackman:
Automated documentation update for [pipecat PR #4501](https://github.com/pipecat-ai/pipecat/pull/4501).

## Changes

- **filter-incomplete-turns.mdx**: Added usage example showing that turn completion works seamlessly with LLM function calls. Documents that function calls automatically finalize the user turn before execution.

## Gaps identified

None - the PR only modified internal implementation details in `src/pipecat/turns/user_turn_completion_mixin.py` and added example files. The fix ensures function calling now works correctly with the filter incomplete turns feature, which has been documented with a new usage example.


markbackman:
Will make manual update.

PR #853, opened 2026-05-29
docs: update for pipecat PR #4592

markbackman:
Automated documentation update for [pipecat PR #4592](https://github.com/pipecat-ai/pipecat/pull/4592).

## Changes

**api-reference/server/utilities/turn-management/filter-incomplete-turns.mdx**

Updated the deprecated parameter descriptions for `filter_incomplete_user_turns` and `user_turn_completion_config` to align with the source code changes in pipecat PR #4592:

- **filter_incomplete_user_turns**: Updated description to lead with the functional explanation (turn-completion markers: ✓, ○, ◐) instead of generic "Enable LLM-based turn completion detection" text. This matches the source docstring which now presents functionality first, followed by the deprecation directive.

- **user_turn_completion_config**: Enhanced description to clarify that this parameter is only used when `filter_incomplete_user_turns` is `True` (the deprecated path), and that the new strategy-based API should use `FilterIncompleteUserTurnStrategies(config=...)` instead.

Both parameters retain their `deprecated` attribute and deprecation notices, consistent with the Sphinx `.. deprecated::` directive in the source.

## Gaps identified

None

### DIFF
--- a/api-reference/server/utilities/turn-management/filter-incomplete-turns.mdx
+++ b/api-reference/server/utilities/turn-management/filter-incomplete-turns.mdx
@@ -101,11 +101,12 @@ The following `LLMUserAggregatorParams` fields predate `FilterIncompleteUserTurn
   default="False"
   deprecated
 >
-  Enable LLM-based turn completion detection. When `True`, the system
-  automatically appends turn completion instructions to the LLM's system prompt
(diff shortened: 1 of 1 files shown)


markbackman:
Unnecessary change. Closing.

PR #859, opened 2026-06-01
Reorganize Learning Pipecat and Fundamentals around the multi-agent path

aconchillo:
## Summary

Surfaces Pipecat's multi-agent features (which were buried in Fundamentals) by promoting the full usage path into **Learning Pipecat**, and reorganizes the remaining Fundamentals section thematically.

### Learning Pipecat: a complete multi-agent arc

The learning path now runs end to end through building and scaling a multi-agent system:

```
… → Multiple LLM Agents → Controlling the UI → Agent Handoff
→ Custom Voices per Agent → Job Coordination → Distributed Agents
→ Proxy Agents → What's Next
```

- **Moved into Learning Pipecat** (from Fundamentals): UIWorker, Custom Voices per Agent, Distributed Agents, Proxy Agents.
- **Renamed `llm-agent-and-tool-decorator` → `multiple-llm-agents`** ("Multiple LLM Agents") with a new intro framing separate agents that each own their own LLM.
- **Renamed the UIWorker page to "Controlling the UI"** (slug stays `ui-worker`), leading with what the agent does rather than the voice-enabling angle.
- **Re-threaded every affected "What's next" card** so the path reads in order.

### Fundamentals

- **Renamed the `Multi-agent` subgroup → `Multi-Agent Architecture`** — after the move it holds exactly the architectural primitives (bus, registry/discovery, bus bridge), and it stays at the bottom as optional reference.
- **Reordered the flat pages thematically**: configure → turn/input control → context → capture → observe/evaluate → telephony → extend.

### Plumbing

- Added redirects from every old `/pipecat/fundamentals/...` path, repointed the `/subagents/...` redirects, and fixed all inbound links (API reference, getting-started, cross-links).
- `docs.json` validated; `mint broken-links` passes clean.


aconchillo:
image
image

markbackman:
This re-org makes sense. `Custom Voices per Agent` feels like the only section that doesn't belong in the learning path. The rest are well suited, teaching a core principle.

aconchillo:
> This re-org makes sense. `Custom Voices per Agent` feels like the only section that doesn't belong in the learning path. The rest are well suited, teaching a core principle.

But it's not a fundamental either, I think.

markbackman:
Maybe just remove and let the example be the guide?

markbackman (approved):

PR #890, opened 2026-06-11
fix(typo): correct env.example to .env.example in quickstart and deployment docs

picsoung:
Hey folks

As I was testing the quickstart, I got stuck at the step where you copy the environment template
 `cp env.example .env` was failing with a "no such file" error, because the actual file is named `.env.example` (with a leading dot).

I found the same typo in a couple of other places too (the Modal deployment guide and the client quickstart).

This PR updates them all to the correct file name.


markbackman (approved):
Nice catch! 🕵️

PR #891, opened 2026-06-11
docs: add available voices list to Mistral TTS page

remisharrock:
## What
- Adds a complete "Available Voices" section to the Mistral TTS documentation page,
  listing all 30 built-in voices available via the Voxtral TTS API
- Renames "With Custom Settings" to "With a Built-in Voice" and adds a separate
  "With a Custom Voice" section to clarify the difference

## Why
The `voice` parameter was documented but with no indication of valid values,
causing errors like `Voice 'neutral_female' not found` (HTTP 404) for users
who guessed descriptive names instead of using the correct slugs or UUIDs.
The distinction between built-in voices and custom (user-created) voices was
also not clear.

## Changes
- Added "Available Voices" section between Settings and Usage, listing all 30
  voices grouped by language (fr_fr, en_gb, en_us) with slug, mood and UUID
- Added curl snippet showing how to fetch the latest voice list from the API
- Renamed "With Custom Settings" to "With a Built-in Voice" with a real slug example
- Added new "With a Custom Voice" section explaining UUID usage for custom voices
- Added a note clarifying that both slugs and UUIDs are accepted

## How voices were verified
Voices retrieved on 2026-06-11 via:
  GET https://api.mistral.ai/v1/audio/voices?limit=200&offset=0

Note: Mistral may add or remove voices over time. The curl snippet is included
so users can always fetch the current list themselves.

### DIFF
--- a/api-reference/server/services/tts/mistral.mdx
+++ b/api-reference/server/services/tts/mistral.mdx
@@ -8,6 +8,7 @@ description: "Text-to-speech service using Mistral's Voxtral TTS API"
 `MistralTTSService` provides text-to-speech synthesis using Mistral's Voxtral TTS API. It uses HTTP streaming with Server-Sent Events to generate PCM-encoded audio at 24kHz.
 
 Key features include:
+
 - Streaming TTS with real-time audio generation
 - Automatic resampling to pipeline sample rate
 - Built-in metrics support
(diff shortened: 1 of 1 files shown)


api-reference/server/services/tts/mistral.mdx
```diff
@@ -84,6 +84,63 @@ Runtime-configurable settings passed via the `settings` constructor argument usi
 | `voice`    | `str`             | `None`                   | Voice identifier. _(Inherited.)_            |
 | `language` | `Language \| str` | `None`                   | Language for synthesis. _(Inherited.)_      |
 
+## Available Voices
```
markbackman, line 88: This information is prone to get stale. I'd prefer to either remove or link out to Mistral's voices list; can you update to one of those?

remisharrock: ha that's the thing, there is no official list of voices, which I find curious but at the same time normal because this french company is growing fast... they are actually recruiting for the "docs" side hehe (Lead Documentation Engineer, Technical Documentation Manager, Product Designer - Documentation...). 
That's why I added a curl snippet showing how to fetch the latest voice list from the API. That's actually how I made that list ! 
So we could give a few examples and say how to get them with curl ? What do you think ?

markbackman: That's very surprising!

Yes, showing how to get a voices list would be better. Thank you.

remisharrock: I did a commit for that

markbackman (approved):
Very nice! Thanks for the update 🙌

PR #914, opened 2026-06-16
Mention uv run pipecat eval for project-installed CLI

aconchillo:
## Summary

The evals docs install the CLI globally (`uv tool install "pipecat-ai[cli]"`) and invoke it as a bare `pipecat eval`, while agents are run with `uv run bot.py`. There was no mention of `uv run pipecat eval`, which is the natural invocation when `pipecat-ai[cli]` is a project dependency rather than a global tool.

## Changes

Added the `uv run pipecat eval` form to the three canonical invocation spots (without rewriting every example, since the global-tool path stays valid):

- `pipecat/evals/overview.mdx` (Requirements) — note that if `pipecat-ai[cli]` is in your project, run the commands with `uv run pipecat eval` (just like `uv run bot.py`).
- `api-reference/cli/eval.mdx` (intro) — same note alongside the existing `python -m pipecat.evals` alternative.
- `pipecat/evals/quickstart.mdx` (Prerequisites) — parenthetical pointing at the `uv run pipecat eval` alternative.

## Verification

`mint broken-links` passes with no broken links.

### DIFF
--- a/api-reference/cli/eval.mdx
+++ b/api-reference/cli/eval.mdx
@@ -5,7 +5,7 @@ description: "Run behavioral evals against a Pipecat agent, individually or as a
 
 Run scenario-based behavioral evals. `pipecat eval run` tests scenarios against an already-running agent; `pipecat eval suite` spawns the agents listed in a manifest and runs their scenarios concurrently. Both exit `0` when everything passes and `1` otherwise.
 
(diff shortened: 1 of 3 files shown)


markbackman (approved):
Makes sense.

PR #928, opened 2026-06-18
Lead with direct functions in Pipecat Flows docs

kompfner:
## What

Pipecat Flows now favors **direct functions** over `FlowsFunctionSchema`-based tool definitions, with `FlowsFunctionSchema` kept as the explicit-schema option (for property control like `enum` / `minimum`/`maximum`). These docs reflect that shift, following the framing in the [base Pipecat function-calling guide](https://docs.pipecat.ai/pipecat/learn/function-calling).

It also reflects [pipecat-flows#283](https://github.com/pipecat-ai/pipecat-flows/pull/283), which made `FlowsFunctionSchema.handler` required.

## Changes

- **Functions guide** — leads with direct functions as the preferred way to define a function; `FlowsFunctionSchema` moved to an "Advanced" section for explicit schema control.
- **Quickstart & State Management guides** — examples define their function as a direct function.
- **Overview** — "Function Types" reframed (node vs. edge by return value); direct function as the normal definition.
- **FlowManager reference** — global-functions example uses a direct function.
- **Types reference** — `FlowsFunctionSchema.handler` marked required; decorator docs updated to `@flows_tool_options` (with `@flows_direct_function` noted as the deprecated alias).

## Follow-up

The quickstart's runnable example (`examples/quickstart/hello_world.py` in pipecat-flows) still uses `FlowsFunctionSchema`; a separate PR will switch it to a direct function to match this guide.

### DIFF
--- a/api-reference/pipecat-flows/overview.mdx
+++ b/api-reference/pipecat-flows/overview.mdx
@@ -81,17 +81,17 @@ uv add "pipecat-ai[daily,openai,deepgram,cartesia,silero]"
 
 ## Function Types
 
+A node's functions either do work within the current node or transition to another node:
+
 ### Node Functions
 
-Execute operations within a single conversation state without switching nodes. Return `(FlowResult, None)`.
+Do work within the current conversation state without switching nodes. Return `(result, None)`.
 
 ### Edge Functions
 
-Create transitions between conversation states, optionally processing data first. Return `(FlowResult, NodeConfig)`.
+Transition to another conversation state, optionally doing work first. Return `(result, next_node)`.
 
-### Direct Functions
-
-Functions passed directly to NodeConfig with automatic metadata extraction from signatures and docstrings. See [`flows_tool_options`](/api-reference/pipecat-flows/types#flows_tool_options-decorator) and [`FlowsDirectFunction`](/api-reference/pipecat-flows/types#flowsdirectfunction).
+Define either kind as a direct function, or — for explicit schema control — with a [`FlowsFunctionSchema`](/api-reference/pipecat-flows/types#flowsfunctionschema). See the [Functions guide](/pipecat-flows/guides/functions).
 
 ## LLM Provider Support
 
--- a/api-reference/pipecat-flows/overview.mdx
+++ b/api-reference/pipecat-flows/overview.mdx
@@ -83,15 +83,10 @@ uv add "pipecat-ai[daily,openai,deepgram,cartesia,silero]"
 
 A node's functions either do work within the current node or transition to another node:
 
-### Node Functions
+- **Node functions** do work within the current conversation state without switching nodes. They return `(result, None)`.
+- **Edge functions** transition to another conversation state, optionally doing work first. They return `(result, next_node)`.
 
-Do work within the current conversation state without switching nodes. Return `(result, None)`.
-
-### Edge Functions
-
-Transition to another conversation state, optionally doing work first. Return `(result, next_node)`.
-
-Define either kind as a direct function, or — for explicit schema control — with a [`FlowsFunctionSchema`](/api-reference/pipecat-flows/types#flowsfunctionschema). See the [Functions guide](/pipecat-flows/guides/functions).
+Define either kind as a direct function, or — for advanced usage — with a [`FlowsFunctionSchema`](/api-reference/pipecat-flows/types#flowsfunctionschema). See the [Functions guide](/pipecat-flows/guides/functions).
 
 ## LLM Provider Support
 
--- a/api-reference/pipecat-flows/types.mdx
+++ b/api-reference/pipecat-flows/types.mdx
@@ -118,7 +118,7 @@ Dataclass for defining function call schemas with Flows-specific properties. Pro
   List of required parameter names from `properties`.
 </ParamField>
 
-<ParamField path="handler" type="FunctionHandler" default="None">
+<ParamField path="handler" type="FunctionHandler" required>
   Function handler to process the function call. Can be a modern handler `(args,
   flow_manager)`, legacy handler `(args)`, or zero-arg handler `()`. Legacy and
   zero-arg handlers are deprecated. The handler returns any JSON-serializable
@@ -276,7 +276,9 @@ config = ContextStrategyConfig(
 
 ## flows_tool_options Decorator
 
-Decorator that configures call options for a Pipecat direct function in Flows.
+Decorator that overrides a function's default call options. It's optional — a
+function works without it; use it only to change `cancel_on_interruption` or
+`timeout_secs`.
 
 ```python
 @flows_tool_options(*, cancel_on_interruption: bool = False, timeout_secs: float | None = None)
@@ -287,16 +289,22 @@ Decorator that configures call options for a Pipecat direct function in Flows.
 | `cancel_on_interruption` | `bool`  | `False` | Whether to cancel the function call when the user interrupts.                             |
 | `timeout_secs`           | `float` | `None`  | Optional per-tool timeout in seconds, overriding the global `function_call_timeout_secs`. |
 
-This decorator is optional; use it to override the defaults for a Flows direct function. Direct functions have their schema automatically extracted from the function signature and docstring. The first parameter must be `flow_manager: FlowManager`, and all other parameters become the function's properties. The docstring provides the function description and parameter descriptions (Google-style).
+A direct function's schema is automatically extracted from its signature and docstring. The first parameter must be `flow_manager: FlowManager`, and all other parameters become the function's properties. The docstring provides the function description and parameter descriptions (Google-style).
 
-Direct functions must return a [`ConsolidatedFunctionResult`](#consolidatedfunctionresult) tuple.
+A direct function must return a [`ConsolidatedFunctionResult`](#consolidatedfunctionresult) tuple.
+
+<Note>
+  `@flows_direct_function` is a deprecated alias of `@flows_tool_options`. It
+  still works but emits a `DeprecationWarning`; use `@flows_tool_options`
+  instead.
+</Note>
 
 ### Example
 
 ```python
 from pipecat_flows import FlowManager, flows_tool_options, ConsolidatedFunctionResult
 
-@flows_tool_options(cancel_on_interruption=False)
+@flows_tool_options(cancel_on_interruption=True)
 async def lookup_order(
     flow_manager: FlowManager, order_id: str
 ) -> ConsolidatedFunctionResult:
@@ -324,7 +332,10 @@ node_config: NodeConfig = {
 ### Deprecated: flows_direct_function
 
 <Warning>
-  **Deprecated.** The `@flows_direct_function` decorator has been renamed to `@flows_tool_options`. The old name still works but emits a `DeprecationWarning` and will be removed in a future version. Update your code to use `@flows_tool_options` instead.
+  **Deprecated.** The `@flows_direct_function` decorator has been renamed to
+  `@flows_tool_options`. The old name still works but emits a
+  `DeprecationWarning` and will be removed in a future version. Update your code
+  to use `@flows_tool_options` instead.
 </Warning>
 
 ## FlowsDirectFunction
--- a/api-reference/pipecat-flows/types.mdx
+++ b/api-reference/pipecat-flows/types.mdx
@@ -210,15 +210,15 @@ TypedDict for configuring actions that execute during node transitions.
 ### Example
 
 ```python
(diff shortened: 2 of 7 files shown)


api-reference/pipecat-flows/overview.mdx
```diff
@@ -81,17 +81,17 @@ uv add "pipecat-ai[daily,openai,deepgram,cartesia,silero]"
 
 ## Function Types
 
+A node's functions either do work within the current node or transition to another node:
```
kompfner, line 84: The colon doesn't look right on the page here.

kompfner: Actually: maybe we should make `Node Functions` and `Edge Functions` bullets, like:

```
- **Node functions** do work...
```

That'd also fix the problem of "Define either kind..." looking like it's under the `Edge Functions` heading, whereas it actually applied to both kinds.

api-reference/pipecat-flows/overview.mdx
```diff
@@ -81,17 +81,17 @@ uv add "pipecat-ai[daily,openai,deepgram,cartesia,silero]"
...
 
 ### Edge Functions
 
-Create transitions between conversation states, optionally processing data first. Return `(FlowResult, NodeConfig)`.
-
-### Direct Functions
+Transition to another conversation state, optionally doing work first. Return `(result, next_node)`.
 
-Functions passed directly to NodeConfig with automatic metadata extraction from signatures and docstrings. See [`flows_direct_function`](/api-reference/pipecat-flows/types#flows_direct_function-decorator) and [`FlowsDirectFunction`](/api-reference/pipecat-flows/types#flowsdirectfunction).
+Define either kind as a direct function, or — for explicit schema control — with a [`FlowsFunctionSchema`](/api-reference/pipecat-flows/types#flowsfunctionschema). See the [Functions guide](/pipecat-flows/guides/functions).
```
kompfner:
```suggestion
Define either kind as a direct function, or — for advanced usage — with a [`FlowsFunctionSchema`](/api-reference/pipecat-flows/types#flowsfunctionschema). See the [Functions guide](/pipecat-flows/guides/functions).
```

api-reference/pipecat-flows/types.mdx
```diff
@@ -274,29 +274,37 @@ config = ContextStrategyConfig(
 )
 ```
 
-## flows_direct_function Decorator
+## flows_tool_options Decorator
 
-Decorator that attaches metadata to a Pipecat direct function for use in Flows.
+Decorator that overrides a function's default call options. It's optional — a
```
kompfner: No need to list the call options here - it'll be obvious directly below.

api-reference/pipecat-flows/types.mdx
```diff
@@ -274,29 +274,37 @@ config = ContextStrategyConfig(
...
 ```
 
 | Parameter                | Type    | Default | Description                                                                               |
 | ------------------------ | ------- | ------- | ----------------------------------------------------------------------------------------- |
 | `cancel_on_interruption` | `bool`  | `False` | Whether to cancel the function call when the user interrupts.                             |
 | `timeout_secs`           | `float` | `None`  | Optional per-tool timeout in seconds, overriding the global `function_call_timeout_secs`. |
 
-Direct functions have their schema automatically extracted from the function signature and docstring. The first parameter must be `flow_manager: FlowManager`, and all other parameters become the function's properties. The docstring provides the function description and parameter descriptions (Google-style).
+A direct function's schema is automatically extracted from its signature and docstring. The first parameter must be `flow_manager: FlowManager`, and all other parameters become the function's properties. The docstring provides the function description and parameter descriptions (Google-style).
```
kompfner: Now that `flows_tool_options` applies to both direct function and function-schema-based ones, it doesn't make sense to define direct functions here. So I think we can remove this paragraph and the next. Keep the `<Note>`, though.

pipecat-flows/guides/functions.mdx
```diff
@@ -19,88 +19,101 @@ For example, if your node's job is to collect a user's favorite color:
 3. The LLM calls the function with the answer
 4. The function processes the data and determines the next node
 
-## Function Definition
+## Defining a Function
 
-Flows provides a universal `FlowsFunctionSchema` that works across all LLM providers:
+The preferred way to define a function is with a **direct function**: a single async function that is _both_ the handler and the schema. Flows auto-derives the function's metadata — name, description, parameter properties (with their descriptions), and which parameters are required — from the function's signature and docstring.
```
kompfner:
```suggestion
The preferred way to define a function is with a **direct function**: a single async function that is _both_ the handler (the code that runs) and the schema (what is advertised to the LLM). Flows auto-derives the function's metadata — name, description, parameter properties (with their descriptions), and which parameters are required — from the function's signature and docstring.
```

pipecat-flows/guides/functions.mdx
```diff
@@ -19,88 +19,101 @@ For example, if your node's job is to collect a user's favorite color:
...
-    required=["color"],
-    handler=record_favorite_color_and_set_next_node,
-    properties={"color": {"type": "string"}},
-)
-```
-
-## Function Handlers
-
-Each function has a corresponding `handler` where you implement your application logic and specify the next node:
+The first parameter is always `flow_manager` (a `FlowManager`); the function's own parameters follow. Document each parameter in a Google-style docstring.
```
kompfner: Maybe we could make `FlowManager` here a link to the `FlowManager` [type documentation](http://localhost:3000/api-reference/pipecat-flows/flow-manager)?

pipecat-flows/guides/functions.mdx
```diff
@@ -19,88 +19,101 @@ For example, if your node's job is to collect a user's favorite color:
...
-    Write to a database, make an API call, etc.
+    Args:
+        color: The user's favorite color.
     """
-    print(f"Your favorite color is: {args['color']}")
-    return args["color"], create_end_node()
+    print(f"Your favorite color is: {color}")
+    return color, create_end_node()
+
+# List the function in a node's `functions`
```
kompfner: Prefer using explicit `NodeConfig()` constructor, like we do in our examples. Feel free to use something like `# ...` to not have to show all fields.

Looks like we use dict-style `NodeConfig` in a few other files, too (and elsewhere in this file). Let's use the `NodeConfig()` constructor wherever we can.

pipecat-flows/guides/functions.mdx
```diff
@@ -19,88 +19,101 @@ For example, if your node's job is to collect a user's favorite color:
...
+- **Next Node**: The `NodeConfig` for Flows to transition to next, or `None`.
 
-Some handlers may not want to transition conversational state, in which case you can return `None` for the next node. Other handlers may _only_ want to transition conversational state without doing other work, in which case you can return `None` for the result.
+A function that doesn't need to change conversational state can return `None` for the next node. One that _only_ changes conversational state, without doing other work, can return `None` for the result.
 
-## Direct Functions
+## Per-Function Call Options
 
-For more concise code, you can optionally use Direct Functions where the function definition and handler are combined in a single function. The function signature and docstring are automatically used to generate the function schema:
+By default, a direct function is not cancelled when the user interrupts, and it uses the LLM service's global timeout. To override either, decorate the function with `@flows_tool_options`. The decorator only attaches call options — the schema is still auto-derived — so decorated functions can stay at module level.
```
kompfner:
```suggestion
By default, a function is cancelled when the user interrupts, and it uses the LLM service's global timeout. To override either, decorate the function with `@flows_tool_options`.
```

kompfner: This suggestion fixes a couple of things:
- per-function call options aren't specific to direct functions
- functions *are* cancelled by default when the user interrupts

pipecat-flows/guides/functions.mdx
```diff
@@ -19,88 +19,101 @@ For example, if your node's job is to collect a user's favorite color:
...
 
 ```python
-async def record_favorite_color(
+from pipecat_flows import flows_tool_options
+
+@flows_tool_options(cancel_on_interruption=True, timeout_secs=10)
```
kompfner: Maybe let's use the `long_running_task` example from the CHANGELOG instead (but keep `timeout_secs`, maybe set to something higher than 10)

pipecat-flows/guides/quickstart.mdx
```diff
@@ -85,19 +77,25 @@ def create_end_node() -> NodeConfig:
   identically. This example uses both styles to show you the options.
 </Tip>
 
-## Write the Function Handler
+## Write the Function
 
-When the LLM calls the function, the handler processes the result and returns the next node:
+When the LLM calls the function, it processes the result and returns the next node. `record_favorite_color` is a **direct function**: its first parameter is `flow_manager`, the rest (here, `color`) become the function's parameters, and Flows derives the schema from the signature and docstring:
```
kompfner:
```suggestion
When the LLM calls the function, it processes the result and returns the next node. `record_favorite_color` is a **direct function**: its first parameter is `flow_manager`, the rest (here, `color`) become the function's parameters, and Flows derives the schema (advertised to the LLM) from the signature and docstring:
```

pipecat-flows/guides/quickstart.mdx
kompfner, line 1: Let's be absolutely certain that this matches the new state of the quickstart: https://github.com/pipecat-ai/pipecat-flows/pull/285

api-reference/pipecat-flows/overview.mdx
kompfner: Fixed — `Node functions` / `Edge functions` are bullets now, so the colon introduces the list.

kompfner: Done — bulleted them, and the "Define either kind…" line now follows the list so it reads as applying to both.

api-reference/pipecat-flows/overview.mdx
kompfner: Applied — changed to "for advanced usage".

api-reference/pipecat-flows/types.mdx
kompfner: Done — dropped the call-options list from the intro (they're in the table right below).

api-reference/pipecat-flows/types.mdx
kompfner: Done — removed the two direct-function paragraphs; kept the `<Note>`.

pipecat-flows/guides/functions.mdx
kompfner: Applied your suggestion (handler = the code that runs; schema = what's advertised to the LLM).

pipecat-flows/guides/functions.mdx
kompfner: Done — `FlowManager` now links to its reference page.

pipecat-flows/guides/functions.mdx
kompfner: Done — using the `NodeConfig()` constructor here and across the other guide/reference snippets (with `# ...` to elide fields). Left the historical migration guide alone.

pipecat-flows/guides/functions.mdx
kompfner: Took the "a function" generalization, but kept **"is not cancelled by default"**: the current Flows default is actually `cancel_on_interruption=False`. It was deliberately flipped `True`→`False` in pipecat-ai/pipecat-flows#237 (fixing #234, where an interrupted transition left the flow stuck), so functions run to completion through interruptions — intentional and Flows-specific (base Pipecat defaults to `True`).

pipecat-flows/guides/functions.mdx
kompfner: Reworked — since the default is already non-interruptible (`cancel_on_interruption=False`, #237), a long-running task doesn't demonstrate an *override*. Switched it to opt *into* cancellation (`cancel_on_interruption=True`) for a turn-specific lookup, matching your point that the example might not fit.

pipecat-flows/guides/quickstart.mdx
kompfner: Applied — "…derives the schema (advertised to the LLM) from the signature and docstring".

pipecat-flows/guides/quickstart.mdx
kompfner: Verified against #285: the quickstart now uses the direct function and the `NodeConfig()` constructor for both nodes, matching `hello_world.py`. I also renamed the example's end node `create_end_node`→`end` there so the node names line up.

markbackman (approved):
Very nice!

PR #934, opened 2026-06-22
docs(evals): document DTMF turns and reorganize the scenarios page

aconchillo:
## Summary

Documents the new evals scenario-format features from the pipecat `aleix/eval-dtmf` work, and reorganizes the scenarios page so the format reads as a coherent whole.

### New features documented
- **DTMF turns (`dtmf:`)**: pressing keypad keys as an alternative to `user:` (mutually exclusive). Covers how a `DTMFAggregator` turns keys into a transcription that drives a full user turn, the `#`-terminator vs idle-timeout flush, valid keys, and the YAML quoting gotcha.
- **Bare `send_after:` delays**: the `event:` anchor is now optional, so `send_after: { delay_ms: N }` is a pure time delay (useful for pacing DTMF keypresses). A zero-delay no-event form is rejected.
- **Events table fix**: `user_transcription` / `user_started_speaking` / `user_stopped_speaking` also fire on DTMF turns in either modality, not just audio mode.

### Reorganization
The page is now four top-level parts:
- **Configuration** (first): all scenario-wide setup, including the text/audio modes overview, `user:` delivery, `judge:`, `factory:`, `context:`, `!include`, and back-to-back / disconnect lifecycle.
- **User turns**: per-turn authoring only, with subsections for the ways a turn drives the agent (`user:`, `dtmf:`, `image:`, `send_after:`).
- **Events** / **Assertions**: kept separate.

This splits the two meanings of `user:` cleanly (the top-level delivery block lives in Configuration, the per-turn utterance in User turns) and cross-links the `eval:` assertion to the judge LLM that decides it.

### Housekeeping
- Removed em dashes from the page's prose.
- Unified the heading/prose style to use a trailing colon whenever naming a YAML key (`eval:`, `text_contains:`, `within_ms:`).

`mint broken-links` passes; prettier-clean.

### DIFF
--- a/pipecat/evals/scenarios.mdx
+++ b/pipecat/evals/scenarios.mdx
@@ -29,11 +29,11 @@ turns: # required: the conversation, in order
 
 Each turn optionally sends a user utterance (`user:`) and lists the events expected in response (`expect:`). Expected events must arrive in the order listed, but the agent may emit other events in between, so you don't have to enumerate everything it does.
 
-The rest of this page is in four parts: [Configuration](#configuration) for scenario-wide setup (modalities, the judge, context, lifecycle), then [Input](#input) for driving each turn, and [Events](#events) and [Assertions](#assertions) for checking what the agent emits.
+The rest of this page is in four parts: [Configuration](#configuration) for scenario-wide setup (modalities, the judge, context, lifecycle), then [User turns](#user-turns) for driving each turn, and [Events](#events) and [Assertions](#assertions) for checking what the agent emits.
 
 ## Configuration
 
-Everything in this section is optional. A scenario with no configuration blocks runs entirely in [text mode](#text-and-audio-modes) with the default judge, which is the fastest way to start. To write the turns themselves, skip ahead to [Input](#input) and [Events](#events).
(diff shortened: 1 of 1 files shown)


pipecat/evals/scenarios.mdx
```diff
@@ -21,146 +21,21 @@ turns: # required: the conversation, in order
...
-  expect:
-    - event: function_call
-      calls:
-        - name: get_current_weather
-          args: { location: "San Francisco" }
-        - name: get_restaurant_recommendation
-    - event: response
-      eval: "describes the weather and recommends a restaurant"
-```
+The rest of this page is in four parts: [Configuration](#configuration) for scenario-wide setup (modalities, the judge, context, lifecycle), then [User turns](#user-turns) for driving each turn, and [Events](#events) and [Assertions](#assertions) for checking what the agent emits.
```
markbackman: I think you should make these cards or something that's more visual. It will break up the page and make this part easier to scan.

markbackman (approved):
Really nice clean up. Just one suggestion.

pipecat/evals/scenarios.mdx
aconchillo: Done!

PR #951, opened 2026-06-26
docs: close gaps found in Kapa source analytics audit

jamsea:
## What

Audited the **top 10 most-cited `docs.pipecat.ai` pages** from a Kapa source analytics export against the actual user questions that cited each page, then fixed the gaps that made Kapa give weak, stale, or contradictory answers. Every change was verified against the `pipecat` source before writing.

## Why

Kapa cites a page each time it answers a user question. Where a highly-cited page was missing a detail, Kapa had to reconstruct answers from GitHub source or Discord (or contradicted itself), which is a signal the docs page should own that answer.

## Changes by page

| Page | Cites | Change |
|---|---|---|
| `stt/deepgram` | 182 | **Fix broken Flux import path** (`...flux` to `...flux.stt`) that threw an `ImportError` in copy-pasted examples |
| `user-turn-strategies` | 438 | Add custom max-turn stop-strategy example, turn-timing timeline, `vad_analyzer` param, 1.x migration note |
| `speech-input` | 326 | Explain why VAD is configured on the aggregator; cross-link noise-reduction filters |
| `function-calling` | 216 | Document `run_in_parallel` / `group_parallel_tools` for duplicate responses with parallel tool calls |
| `context-management` | 181 | Silent add (`run_llm=False`), `LLMMessagesTransformFrame`, `TTSSpeakFrame(append_to_context=True)` link |
| `text-to-speech` | 179 | State the `append_to_context` default (`True`, as of v1.4.0); resolves contradictory answers |
| `choosing-a-transport` | 175 | Clarify LiveKit is server-side only (no client transport); add telephony WebSocket use case |
| `migration-1.0` | 168 | Warn that removed params are silently ignored; add a concrete interruptions mapping |
| `smart-turn-overview` | 168 | Disambiguate the two `stop_secs` values (VAD 0.2 vs SmartTurnParams 3.0) |
| `pipeline-termination` | 158 | Frame-terminology table, complete the end-call example (`result_callback` + registration), `on_pipeline_finished` cleanup, max call duration, dangling-tasks troubleshooting |

## Notes

- Source-verified against `../pipecat` for every API name, import path, default value, and deprecation.
- One known-stale value was left out of scope: `smart-turn-overview` lists `pre_speech_ms` default as `0.0` but source says `500`. Flagging for a separate fix.
- Formatted with Prettier. Internal links confirmed to resolve.

### DIFF
--- a/client/concepts/choosing-a-transport.mdx
+++ b/client/concepts/choosing-a-transport.mdx
@@ -10,19 +10,39 @@ The choice of transport has one important constraint: **the client transport and
 ## Available transports
 
 <CardGroup cols={2}>
-  <Card title="SmallWebRTC" icon="link" href="/api-reference/client/js/transports/small-webrtc">
+  <Card
+    title="SmallWebRTC"
+    icon="link"
+    href="/api-reference/client/js/transports/small-webrtc"
+  >
     Serverless peer-to-peer WebRTC. No third-party account needed.
   </Card>
-  <Card title="Daily" icon="tower-broadcast" href="/api-reference/client/js/transports/daily">
+  <Card
+    title="Daily"
+    icon="tower-broadcast"
+    href="/api-reference/client/js/transports/daily"
+  >
     WebRTC via Daily's global infrastructure. Recommended for production.
   </Card>
-  <Card title="WebSocket" icon="plug" href="/api-reference/client/js/transports/websocket">
+  <Card
+    title="WebSocket"
+    icon="plug"
+    href="/api-reference/client/js/transports/websocket"
+  >
     Direct WebSocket connection. For server-to-server setups only.
   </Card>
-  <Card title="Gemini Live" icon="google" href="/api-reference/client/js/transports/gemini">
+  <Card
+    title="Gemini Live"
+    icon="google"
+    href="/api-reference/client/js/transports/gemini"
+  >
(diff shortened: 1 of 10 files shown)


client/concepts/choosing-a-transport.mdx
```diff
@@ -127,11 +154,22 @@ The WebSocket transport connects to a WebSocket server rather than using WebRTC.
...
+  for development and demos, but **not for production** — use a server-side
+  Pipecat pipeline with `DailyTransport` or `SmallWebRTCTransport` to keep API
+  keys secure.
 </Warning>
 
 ---
 
+### What about LiveKit?
+
+Pipecat ships a [LiveKit transport for the **server**](/api-reference/server/services/transport/livekit), so your bot can join a LiveKit room. There is **no** official Pipecat **client** transport for LiveKit (JS/React, React Native, or otherwise).
```
jamsea, line 167: @markbackman this is a direct response to people asking the AI chatbots about Livekit. We do have an open PR here to make a transport: https://github.com/pipecat-ai/pipecat-client-web-transports/pull/86

Might be worth merging in the PR and deleting this. But not sure how much of a lift that is for engineering. While we do get people asking the AI chatbot we don't have anyone on paid support asking about this so it might be lower in priority.

markbackman (approved):

PR #953, opened 2026-06-26
GoogleVertexLLMService import fix

grvbrk:
Updated the import path for GoogleVertexLLMService class. 
Earlier import: from pipecat.services.google import GoogleVertexLLMService
Current import: from pipecat.services.google.vertex.llm import GoogleVertexLLMService


markbackman (approved):
LGTM! Thanks 🙇

PR #962, opened 2026-06-29
Add Latitude to OpenTelemetry exporter options

guillemwilly:
Adds a **Latitude** subsection to the Exporter Options on the OpenTelemetry page.

[Latitude](https://latitude.so) is an open-source LLM observability and evaluation platform that ingests OTLP traces. Like the existing Langfuse HTTP OTLP usage, it needs no Pipecat-specific package: configure the HTTP OTLP exporter with Latitude's ingestion endpoint and `Authorization` / `X-Latitude-Project` headers. Single-file docs change with a runnable snippet.


markbackman (commented):
Rather than submit just a docs change, I'd encourage you to add an example here:
https://github.com/pipecat-ai/pipecat-examples/tree/main/open-telemetry

And then we can link to the example from this section. Does that make sense?

PR #965, opened 2026-06-29
docs(assemblyai): focus STT docs on universal-3-5-pro

dlange-aai:
## What

Simplifies the AssemblyAI STT reference so a reader lands on the page, sees one model — `universal-3-5-pro` — and isn't asked to weigh alternatives.

- Removes all references to `universal-streaming-english` and `universal-streaming-multilingual`.
- Drops the two parameters that only function on those models: `language_detection` and `format_turns`.
- Removes the now-redundant "(`universal-3-5-pro` only)" qualifiers that were there to contrast against the other models.
- Sets `model="universal-3-5-pro"` explicitly in every code snippet.

## Note on "default"

I intentionally avoided asserting in prose that `universal-3-5-pro` is *the default*. The latest published SDK (**v1.4.0**) still defaults to the equivalent `u3-rt-pro` alias; the default only flips to `universal-3-5-pro` in unreleased `main` (commit `ceb8bfff9`). `universal-3-5-pro` is a supported model on v1.4.0 with the full U3 Pro feature set, so the explicit `model="universal-3-5-pro"` in the snippets is correct on the current release and forward-compatible.

## Source divergence (intentional)

The Pipecat source still lists `universal-streaming-english` / `universal-streaming-multilingual` in the `speech_model` literal. Dropping them from the docs is a deliberate de-emphasis to reduce reader confusion, not a claim that they were removed from the SDK.


markbackman (approved):
LGTM!

PR #976, opened 2026-07-02
docs: add note on filtering prompt-elicited inline tags (e.g. <thinking>) before TTS

scttbnsn:
## Summary

Follow-up to pipecat-ai/pipecat#4901. When a system prompt asks the LLM to reason inside inline `<thinking>...</thinking>` tags with extended thinking off, the reasoning streams back as plain text and gets spoken by TTS. @markbackman's investigation there concluded this belongs in docs rather than provider code: prefer native extended thinking, and strip deliberately-elicited inline tags at the text layer. This adds that note.

## What changed

- New "Removing Custom Inline Tags" section in `pipecat/learn/text-to-speech.mdx` with the `PatternPairAggregator` + `MatchAction.REMOVE` snippet from the issue, plus a Tip pointing at native extended thinking as the preferred path when the goal is genuine reasoning.
- A short bullet in the Notes of `api-reference/server/services/llm/anthropic.mdx` cross-linking the new section, since that's where someone debugging spoken thinking text with Anthropic looks first.

## Verification

- Snippet imports and behavior checked against pipecat main: `LLMTextProcessor(text_aggregator=...)`, `add_pattern(..., action=MatchAction.REMOVE)`, and `LLMThoughtTextFrame` routing in `AnthropicLLMService`.
- `npx prettier` clean on both files.


api-reference/server/services/llm/anthropic.mdx
```diff
@@ -201,6 +201,7 @@ await worker.queue_frame(
 
 - **Prompt caching**: When `enable_prompt_caching` is enabled, Anthropic caches repeated context to reduce costs. Cache control markers are automatically added to the most recent user messages. This is most effective for conversations with large system prompts or long conversation histories.
 - **Extended thinking**: Enabling thinking increases response quality for complex tasks but adds latency. When `type="enabled"`, you must provide a `budget_tokens` value (minimum 1024 with current models). Extended thinking is disabled by default.
+- **Prompt-elicited `<thinking>` tags**: If your system prompt asks the model to reason inside inline tags rather than enabling extended thinking, that reasoning is ordinary text and will be spoken by TTS. Prefer the `thinking` parameter; for inline tags you deliberately keep, see [Removing Custom Inline Tags](/pipecat/learn/text-to-speech#removing-custom-inline-tags).
```
markbackman: This note makes sense.

Rather than adding a new subsection to the learning guides, it might make sense to just point the developer directly to the `PatternPairAggregator`. In the `PatternPairAggregator`, we can add a new, generic section about removing tags, which we can link to. Something like this would do the trick:
```
### Removing Tagged Content

To drop content from the text stream entirely, register a pattern with `MatchAction.REMOVE`. The tags and everything between them are removed before reaching downstream processors — nothing is spoken by TTS and nothing lands in the conversation context. This is useful when your prompt elicits inline tags whose content is not meant for the user, such as reasoning tags (e.g., `<thinking>...</thinking>`) or annotations intended for other processors:

```python
from pipecat.processors.aggregators.llm_text_processor import LLMTextProcessor
from pipecat.utils.text.pattern_pair_aggregator import MatchAction, PatternPairAggregator

pattern_aggregator = PatternPairAggregator()
pattern_aggregator.add_pattern(
    type="thinking",
    start_pattern="<thinking>",
    end_pattern="</thinking>",
    action=MatchAction.REMOVE,
)

# Set the aggregator on an LLMTextProcessor
llm_text_processor = LLMTextProcessor(text_aggregator=pattern_aggregator)

# add the llm_text_processor to your pipeline after the llm and before the tts
# llm -> llm_text_processor -> tts
```

Because this filters the text stream itself, it works with any LLM provider and any custom inline tag.
```

File: pipecat/learn/text-to-speech.mdx [hunk 2]
```diff
@@ -173,6 +173,41 @@ tts = CartesiaTTSService(
 # llm -> llm_text_processor -> tts
 ```
 
+### Removing Custom Inline Tags
```
markbackman: From the other comment, I think we'll want to remove this.

api-reference/server/services/llm/anthropic.mdx
scttbnsn: Done in b117685. Used your text as a new "Removing Tagged Content" example on the PatternPairAggregator page, dropped the learn-guide section, and pointed the anthropic note at the new anchor.

pipecat/learn/text-to-speech.mdx
scttbnsn: Removed.

markbackman (approved):
LGTM! Thanks for taking care of this 🙇

PR #982, opened 2026-07-06
Add NLTK preloading to local deployment example

jaydamani:
Quick PR based on discussion in https://github.com/pipecat-ai/pipecat/issues/4963


markbackman (approved):
LGTM! Thanks for the update.

PR #991, opened 2026-07-14
Add Gnani (STT and TTS) community integrations

Gnani-AI-Mintlify:
## Gnani

Adds **Gnani** to the Community Integrations page under both **Speech-to-Text** and **Text-to-Speech**.

Gnani Vachana is a production speech AI platform for Indian languages, providing high-accuracy STT and low-latency TTS across REST, SSE, and WebSocket. The integration includes `GnaniHttpSTTService`, `GnaniSTTService`, `GnaniHttpTTSService`, `GnaniSSETTSService`, and `GnaniTTSService`. WebSocket STT supports 10+ Indian languages at 8/16/44.1/48 kHz, while TTS offers four voices (Pranav, Kaveri, Shubhra, and Deepak) with interruptible WebSocket streaming.

Tested with the latest Pipecat main branch (v1.5.0+).

**Repository:** https://github.com/Gnani-AI-Mintlify/pipecat-gnani
**PyPI:** https://pypi.org/project/pipecat-gnani/
**Core SDK:** https://pypi.org/project/gnani-vachana/
**Demo Video:** https://go.screenpal.com/watch/cOi2nynUax5
**Maintainers:** [(Gnani.ai)](https://github.com/Gnani-AI-Mintlify)

### DIFF
--- a/api-reference/server/services/stt/gnani.mdx
+++ b/api-reference/server/services/stt/gnani.mdx
@@ -15,8 +15,7 @@ import { CommunityMaintained } from "/snippets/community-maintained.mdx";
 
 `GnaniHttpSTTService` and `GnaniSTTService` transcribe speech using
 [Gnani Vachana](https://gnani.ai/), a production speech AI platform for 10+
-Indian languages with real-time streaming, multilingual transcription, and
-code-switching support.
+Indian languages with real-time streaming and multilingual transcription.
 
 Use `GnaniHttpSTTService` for REST-based file transcription of VAD-segmented
 utterances, or `GnaniSTTService` for low-latency WebSocket streaming with
@@ -154,7 +153,6 @@ Runtime-configurable settings passed via the `settings` constructor argument.
 | `language`            | `Language` | `Language.EN_IN` | Recognition language (BCP-47, e.g. `Language.HI_IN`).                       |
 | `format`              | `str`      | `None`           | `verbatim` for raw output, `transcribe` for ITN-normalized output.          |
 | `itn_native_numerals` | `bool`     | `None`           | When `format=transcribe`, render digits in native script.                     |
-| `preferred_language`  | `str`      | `None`           | Force single-language model when multiple languages are specified.          |
 
 **`GnaniSTTService.Settings`**
 
--- a/api-reference/server/services/stt/gnani.mdx
+++ b/api-reference/server/services/stt/gnani.mdx
@@ -1,0 +1,247 @@
+---
+title: "Gnani"
+description: "Speech-to-text services for Gnani Vachana — REST and WebSocket streaming for Indian languages"
+---
+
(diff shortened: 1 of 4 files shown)


api-reference/server/services/stt/gnani.mdx
```diff
@@ -0,0 +1,246 @@
...
+
+<CommunityMaintained
+  maintainer="Gnani-AI-Mintlify"
+  maintainerUrl="https://github.com/Gnani-AI-Mintlify"
+  repo="https://github.com/Gnani-AI-Mintlify/pipecat-gnani"
+/>
+
+## Overview
+
+`GnaniHttpSTTService` and `GnaniSTTService` transcribe speech using
```
markbackman, line 16: Would it be simpler to just add support for the streaming services? Or, are there use cases where the HTTP API is better suited for voice agents?

Same question for the TTS services.

docs.json
```diff
@@ -1857,6 +1859,10 @@
       "source": "/server/services/stt/gladia",
       "destination": "/api-reference/server/services/stt/gladia"
     },
+    {
```
markbackman: No need for a redirect since these were just added.

docs.json
```diff
@@ -2017,6 +2023,10 @@
       "source": "/server/services/tts/fish",
       "destination": "/api-reference/server/services/tts/fish"
     },
+    {
```
markbackman: Same here, remove.

markbackman (commented):
Before moving forward, I think it's worth considering which APIs to expose to Pipecat. Do you want to maintain HTTP and Websocket endpoints? Pipecat is for realtime / voice use cases, so streaming services are almost always preferred.

docs.json
Gnani-AI-Mintlify: sure , will remove those

api-reference/server/services/stt/gnani.mdx
Gnani-AI-Mintlify: Hi @markbackman , For STT, we already have a streaming service in place, and for TTS we have an SSE-based output streaming service, so the streaming path is fully supported on both sides.
the REST-based approach is more economical. Our latency on the REST path is also good, making it well-suited to voice agents.

markbackman: Makes sense. Include services that are best suited for realtime.

Gnani-AI-Mintlify: @markbackman Thanks for the review! We'd prefer to keep both the realtime and HTTP paths if possible. A significant part of our user base is in India, where cost is a major consideration, so the HTTP option is valuable for us. Our model has also been benchmarked with very low latency, so the tradeoff there is minimal.
Could you help move this PR forward when you get a chance? Happy to make any changes you'd like before merging.

markbackman (approved):
LGTM! Thanks for adding your services 🙌

PR #824, opened 2026-05-14
Expand self-hosting deployment docs

mattshep:
Rewrite the deployment section to frame self-hosting as a real path alongside Pipecat Cloud, with concrete patterns and tradeoffs.

- Rewrite overview.mdx around the dev-runner vs. production-hosting split
- Rewrite pattern.mdx to teach the current bot(runner_args) + runner.run shape, replacing the obsolete subprocess bot_runner pattern
- New self-hosting.mdx bridges dev to prod and enumerates the answer space: dev runner in prod, hand-rolled dispatcher, managed runtime, Pipecat Cloud — plus cross-cutting concerns (auth, secrets, image discipline, lifecycle, observability, networking)
- New patterns/ subsection with three production shapes pointing at existing examples: vm-per-session (flyio-example), warm-pool-subprocess (instant-voice), managed-runtime (aws-agentcore-webrtc)
- New telephony-in-production.mdx covers webhook-driven dispatch, egress IPs, codec/timing gotchas, dial-out and transfers
- Cross-link the Fly, Modal, and Cerebrium platform pages to the matching patterns
- Modernize the broken inline code in cerebrium.mdx: add DeepgramSTT, drop the non-existent VADProcessor and duplicate aggregator, fix the blocking time.sleep, use LLMRunFrame and context.add_message, correct DAILY_TOKEN/CARTERSIA_API_KEY env var names
- Add the new pages to docs.json nav and redirects

PR #995, opened 2026-07-15
docs: position the development runner as not for production use

mattshep:
## What & why

We've been receiving security advisories for the development runner (`pipecat.runner.run`) — most recently an unauthenticated dial-in-webhook endpoint that lets a remote caller spawn a bot and create a paid Daily room at the operator's expense. Internally we've decided **not** to harden the dev runner toward production parity; it exists to streamline local development. Our answer for production is a proper dispatcher, a managed runtime, or Pipecat Cloud.

This updates the deployment docs to state that boundary clearly, rather than presenting the runner as a viable production target (as PR #824 did). It aligns with the maintainer position on the advisory: *"this is a dev runner; NOT meant for production use."*

## Changes

- **`self-hosting.mdx`** — Pull the runner out of the numbered options. Lead with a `<Warning>` stating it's not built or supported for production (unauthenticated/unsigned/unthrottled endpoints; a single request can create paid resources; SPOF). The two real options become **Build your own dispatcher** and **Use a managed runtime**. Add a new **"Exposing the runner for prototyping"** section — the minimum reverse-proxy smarts (auth on `/start`, rate limiting, TLS), explicitly capped at prototyping/alpha, plus the two things it doesn't solve (transport webhooks have a different caller than `/start`; lifecycle/capacity/isolation untouched).
- **`overview.mdx`** — Drop "the runner directly" from the list of things that can serve production session-start traffic.
- **`running-bots-locally.mdx`** — Replace the "genuinely a fine production target" closing note with a clear dev-tool-not-a-production-server boundary. Local-dev happy path is otherwise untouched.

Deliberately **no** links to specific security advisories or CVE IDs — citing them invites a "just patch these and it's fine" read, which reintroduces the "sort-of production-ready" framing we're trying to kill.

## Follow-ups (not in this PR)

- The `--host 0.0.0.0` containerizing example in `running-bots-locally.mdx` still binds the runner to all interfaces; left as-is to keep scope tight, but it's the one remaining spot that soft-contradicts the new posture.
- If/when the runner gains a startup console disclaimer, a one-line reinforcement can be added to the warning.


pipecat/deployment/running-bots-locally.mdx
```diff
@@ -125,6 +125,8 @@ CMD ["python", "bot.py", "--host", "0.0.0.0"]
 
 For production-grade images, Pipecat Cloud's [base image](https://github.com/daily-co/pipecat-cloud-images) (`dailyco/pipecat-base`) is what most of the [examples](https://github.com/pipecat-ai/pipecat-examples) build from. It's also a reasonable starting point for self-hosted deployments — it provides a sensible runtime environment for a bot file, and you can use it without using PCC.
 
-## When the development runner isn't enough
+## The runner is a development tool, not a production server
```
filipi87, line 110: I wonder if we should add a callout for this right at the beginning of this page and link to this section.

What do you think ?

filipi87 (approved):
@mattshep, I think this is great. 

It clearly explains what the runner can and cannot do, as well as the alternatives if you still want to use it for a proof of concept.

🚀🎉

markbackman (approved):
Nice! Thanks for the update.

pipecat/deployment/running-bots-locally.mdx
kompfner, line 110: The section above this one—"Containerizing for deployment"—feels like it already leads the reader down the path of deploying the runner to production. When the reader then gets to this section, it's a bit of whiplash: "hey, I thought you just suggested I *could* deploy this runner..."

So it seems to me that a bit more restructuring is in order for this page. Actually, I wonder if "Containerizing for deployment" doesn't even really belong on a page called "Running bots locally" in the first place...the information in it might more naturally live in the "self-hosting" page.

pipecat/deployment/self-hosting.mdx
```diff
@@ -1,25 +1,33 @@
...
+## What plays the runner's role
 
-A practical upper bound per host is "however many concurrent bot subprocesses one host can sustain", which depends heavily on what your pipeline does. CPU-light pipelines that mostly proxy between transport and third-party services can pack many sessions per host; pipelines with local STT/TTS, custom models, or significant VAD work will saturate sooner.
+The two options below span the range from "you run everything yourself" to "a managed runtime runs the bot for you". The first is self-hosting in the literal sense; the second is an alternative to it. They're walked through together because the decision space is the same one ("what serves session-start and runs the bot?") and most teams evaluating self-hosting are also weighing whether they want to be in that business at all.
```
kompfner: Nit: "span the range of" would work better if there were more than two options. Maybe something more like:

```
The two options can below can roughly be summarized as "run everything yourself" and "let a managed runtime do it for you".
```

kompfner: Also: "They're walked through together because..." would read more smoothly as "Both options are presented together here because..."

pipecat/deployment/self-hosting.mdx
```diff
@@ -28,21 +36,36 @@ You write the HTTP service that receives session-start requests, and you decide
...
 
-If you're considering self-hosting partly because you already have an ops team and existing infrastructure to leverage, options 1–2 are probably the right starting point. If you'd rather not build any of that surface, option 3 is the path designed to skip it.
+If you're considering self-hosting partly because you already have an ops team and existing infrastructure to leverage, Option 1 is probably the right starting point. If you'd rather not build any of that surface, Option 2 is the path designed to skip it.
+
+## Exposing the runner for prototyping
+
+Sometimes you need something reachable before you've built either option above — a demo, an alpha test, a prototype a handful of people will hit. If you're going to put the development runner on a public address at all, treat it as **prototyping-only** and put a reverse proxy in front of it that adds, at minimum:
```
kompfner: I wonder if we can be a *smidge* softer here, with something like "you can do nothing, but be warned". Like:

```
If you're going to put the development runner on a public address at all, treat it as **prototyping-only**. We recommend putting a reverse proxy in front of it that adds, at a minimum:
```

pipecat/deployment/self-hosting.mdx
```diff
@@ -28,21 +36,36 @@ You write the HTTP service that receives session-start requests, and you decide
...
+
+- **Authentication on `POST /start`** — the runner accepts start requests from anyone. A shared secret or bearer token checked at the proxy keeps strangers from spawning bots (and paid rooms) on your account.
+- **Rate limiting** — the runner has no backpressure. A per-IP and global request cap at the proxy bounds both load and cost if the address is discovered.
+- **TLS termination** — the runner speaks plain HTTP.
+
+Two things this does _not_ solve, which is why it stays prototyping-only:
```
kompfner: I would tweak this wording to avoid conveying "if you solve for this very specific list of things, then it *will* be production-ready". Maybe something like:

```
Notably, the above does _not_ address the following:
```

kompfner (commented):
A few suggestions, but overall this is a nice clarification of our recommendations!

pipecat/deployment/self-hosting.mdx
```diff
@@ -1,25 +1,33 @@
 ---
 title: "Self-hosting in production"
```
kompfner: Question: should we broaden the framing of this page to "Running bots in production" instead of "Self-hosting in production"? Only one of the two options presented on this page can be described as "self-hosting".

kompfner: Then the next section—"Self-hosting patterns"—should be renamed to "Production hosting patterns" or something, accordingly, too.

mattshep:
Going through all the review notes, I'll commit some updates in a bit.

mattshep:
All review feedback should now be addressed. @markbackman you might want to take a second look as well as I renamed a couple of pages.

pipecat/deployment/running-bots-in-production.mdx
kompfner, line 1: After the title change, should the name of the file change accordingly?

kompfner: (If so: maybe we'd need to do something to ensure that old links continue to work)

pipecat/deployment/self-hosting.mdx
```diff
@@ -1,25 +1,34 @@
...
+## What plays the runner's role
 
-A practical upper bound per host is "however many concurrent bot subprocesses one host can sustain", which depends heavily on what your pipeline does. CPU-light pipelines that mostly proxy between transport and third-party services can pack many sessions per host; pipelines with local STT/TTS, custom models, or significant VAD work will saturate sooner.
+The two options below can roughly be summarized as "run everything yourself" and "let a managed runtime do it for you". The first is self-hosting in the literal sense; the second is an alternative to it. Both options are presented together here because the decision is the same one ("what serves session-start and runs the bot?"), and most teams evaluating self-hosting are also weighing whether they want to be in that business at all.
```
kompfner: Nit: now that this page isn't titled "self-hosting", we don't need to sort of apologize for one of the options not technically being self-hosting.


```suggestion
The two options below can roughly be summarized as "run everything yourself" and "let a managed runtime do it for you".
```

pipecat/deployment/self-hosting.mdx
```diff
@@ -1,25 +1,34 @@
...
+  and it's a single process with no lifecycle management.
 
-## What plays the runner's role
-
-The options below sit on a spectrum from "you run everything yourself" to "a managed runtime runs the bot for you". Strictly speaking, only the first two are self-hosting in the literal sense — the third is an alternative to self-hosting. They're walked through together because the decision space is the same one ("what serves session-start and runs the bot?") and most teams evaluating self-hosting are also weighing whether they want to be in that business at all.
+Running bots in production means putting one of the options below in front of
+your bots — not exposing the runner to real traffic. If you're only
+prototyping, [Exposing the runner for
+prototyping](#exposing-the-runner-for-prototyping) covers the minimum you'd
+want in front of it, and where that stops being enough.
```
kompfner:
```suggestion
Running bots in production means choosing one of the below options — not exposing the runner to real traffic. If you're only
prototyping, [Exposing the runner for
prototyping](#exposing-the-runner-for-prototyping) covers the minimum you'd
want in front of it, and where that stops being enough.
```

pipecat/deployment/self-hosting.mdx
```diff
@@ -28,29 +37,69 @@ You write the HTTP service that receives session-start requests, and you decide
...
 
 [Pipecat Cloud](/pipecat-cloud/introduction) is one example: Daily's managed runtime, purpose-built for Pipecat. The development runner's session-start API is deliberately shaped the same way as PCC's (`POST /start`, `/sessions/{id}/...`), which means a bot file and client built against the runner work against PCC unchanged. Other managed runtimes — like AWS Bedrock AgentCore — are the same shape, with their own platform-specific tradeoffs.
 
-If you're considering self-hosting partly because you already have an ops team and existing infrastructure to leverage, options 1–2 are probably the right starting point. If you'd rather not build any of that surface, option 3 is the path designed to skip it.
+If you're considering self-hosting partly because you already have an ops team and existing infrastructure to leverage, Option 1 is probably the right starting point. If you'd rather not build any of that surface, Option 2 is the path designed to skip it.
```
kompfner: I'd delete this holdover paragraph—there's only 1 self-hosting option presented now.

pipecat/deployment/self-hosting.mdx
```diff
@@ -28,29 +37,69 @@ You write the HTTP service that receives session-start requests, and you decide
...
+Notably, none of the above addresses the following — which is why this stays prototyping-only:
+
+- **The transport webhooks have a different caller than `POST /start`.** Endpoints like the Daily dial-in webhook are invoked by the transport provider, not your app, so a single proxy secret doesn't fit them — they'd need provider-appropriate verification (signature checks, IP allowlists) that the runner's dial-in webhook doesn't perform. Enabling a telephony transport exposes these endpoints too.
+- **None of this addresses lifecycle, capacity, or isolation.** You still have a single point of failure, no graceful drain, and no way to scale past one host.
+
+If you find yourself building this out, that's the signal you've outgrown the runner — move to Option 1 or Option 2.
 
 ## What you'll need to address either way
 
 Whichever option you pick, a handful of concerns show up in production. None of them have a single right answer; the goal here is just to name them so they don't surprise you.
```
kompfner: Shoot, sorry, just noticed this. This section ("what you'll need to address either way") doesn't make all that much sense anymore, now that we're framing the choice as between self-hosting and using a managed runner (as opposed to various self-hosting options).

The minimal change might be to reframe this section as something like "Self-hosting considerations" (with wording adjusted accordingly throughout)—these considerations only apply if you go with option 1.

kompfner:
So sorry, @mattshep, just noticed that we still have a bunch of language in the "running in production" guide that really only made sense when the guide was focused on self-hosting and only presented the managed option as a bonus. Now that the guide is broadly focused on how to run bots in production, all the stale language should be expunged or updated accordingly. Should've noticed this on my first pass through!

mattshep:
Thanks @kompfner — the latest two pushes close out this round:

- **"choosing one of the options below"** (was "putting one in front of your bots") — agreed, the managed-runtime option isn't something you put in front of a bot.
- **Dropped the "self-hosting in the literal sense" aside** now that the page isn't titled self-hosting.
- **Removed the Option 1-vs-2 chooser paragraph** — holdover from the old three-option structure.
- **Reframed "What you'll need to address either way" → "Self-hosting considerations"** — these are Option 1's to own; a managed runtime handles most of them, with a note that a couple (bot images, region selection) apply either way.
- **File rename:** `self-hosting.mdx` → `running-bots-in-production.mdx`, with the nav path and all internal links updated. Redirects added from both the current slug (`/pipecat/deployment/self-hosting`) and the legacy `/deployment/self-hosting` path, so existing links keep working — the legacy one is repointed straight at the new slug to avoid a redirect chain.

One correction worth surfacing from the earlier security wording: I scoped the "unauthenticated" claim to `POST /start` and the Daily dial-in webhook specifically, rather than a blanket "no request signing." Checking the runner source, it *does* sign its WebSocket session tokens (per-process HMAC-SHA256, single-use, short-lived) and *does* verify the WhatsApp webhook's `x-hub-signature-256`. So the genuinely open surface is `/start` plus the telephony/dial-in webhooks — not every endpoint.

pipecat/deployment/self-hosting.mdx
kompfner: This suggestion still stands.

kompfner (approved):
This looks great! Thanks for the tweaks. I've resolved all my feedback comments except for 1, which I believe still applies. Going ahead and putting my stamp of approval on it. No need to circle back.

pipecat/deployment/self-hosting.mdx
mattshep: It was applied, I think the file rename might be skewing your view.

markbackman (approved):
Nice improvements! Thanks for the help @mattshep 🙌

PR #1003, opened 2026-07-20
Add Quickdial (community TTS + STT integration)

roshanpuru:
This adds **Quickdial** as a community-maintained TTS + STT integration, following [COMMUNITY_INTEGRATIONS.md](https://github.com/pipecat-ai/pipecat/blob/main/COMMUNITY_INTEGRATIONS.md).

Quickdial is a CPU-optimized, real-time TTS/STT API (no GPU, priced per character). The plugin is published and working:

- Package: [`pipecat-quickdial`](https://pypi.org/project/pipecat-quickdial/) on PyPI
- Source: https://github.com/samay-ai/pipecat-quickdial
- `QuickdialTTSService` (extends `TTSService`) and `QuickdialSTTService` (extends `SegmentedSTTService`)

### Changes
- `api-reference/server/services/tts/quickdial.mdx` — new TTS service page (uses the `CommunityMaintained` badge)
- `api-reference/server/services/stt/quickdial.mdx` — new STT service page
- `api-reference/server/services/supported-services.mdx` — added a `Community` row to the TTS and STT tables
- `docs.json` — registered both pages in the TTS and STT nav groups

The integration is maintained by us (Quickdial AI) in our own repo, per the community track. Happy to adjust anything to match your conventions — thanks!

PR #1008, opened 2026-07-21
docs(quickdial): refresh copy + nudge rebuild of the two pages

roshanpuru:
Small follow-up to #1003.

**Changes** (both `api-reference/server/services/{tts,stt}/quickdial.mdx`):
- Refresh the free-tier wording: `1000 free credits` → `100,000 characters free` (matches Quickdial's current sign-up allowance).
- TTS page: simplify the overview/description wording (drop the compute-substrate detail), keeping it focused on the real-time streaming + per-character pricing.

**Heads-up on a rendering issue:** since #1003 merged, the two pages 404 on the live docs (`/api-reference/server/services/tts/quickdial` and `/stt/quickdial`), even though the `.mdx` files, `docs.json` nav entries, and `supported-services` references are all on `main` and every sibling page renders fine. Re-touching these files here should re-trigger Mintlify's build for them — but if they still 404 after merge, it may be a build/cache issue worth a look on your side. Thanks!

### DIFF
--- a/api-reference/server/services/stt/quickdial.mdx
+++ b/api-reference/server/services/stt/quickdial.mdx
@@ -51,7 +51,7 @@ uv add pipecat-quickdial
 
 ## Prerequisites
 
-1. **Quickdial account** — sign up at [web.quickdial.ai](https://web.quickdial.ai) (1000 free credits, no card required).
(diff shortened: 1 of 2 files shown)


markbackman (approved):
👍

markbackman:
> Heads-up on a rendering issue:

Docs needed to be released, which has now happened. I see the docs live now.

PR #1014, opened 2026-07-24
docs: update for Gnani TTS

Gnani-AI-Mintlify:
- Updated dependency version for the Core Gnani Vachana Python SDK from >= 0.7.3 to >= 0.7.9 in both STT and TTS documentation.
- Revised TTS model identifier from "vachana-voice-v3" to "timbre-v2.0" and added details for the new "timbre-v2.5" model, including available voices and language parameters.
- Enhanced parameter descriptions and added examples for using the new TTS model with language support.


markbackman (approved):
LGTM! This will be included in the next docs release, which happens after the Pipecat release. (estimated for today)

PR #1024, opened 2026-07-28
Document Pipecat Cloud session recordings (T-2978)

jamsea:
## Why

A customer asked three questions about the audio recording on the Pipecat Cloud session detail view: where in the chain the audio is captured, where it is stored and how to control its lifecycle for compliance, and what setup their agent needs. None of it was documented. The only page they could find was Logging and Observability, which does not mention recordings at all.

Answering the ticket needed a source read and a log trace. That is the gap this closes.

## What this adds

A new **Session Recordings** page under Pipecat Cloud > Fundamentals, plus a pointer to it from the Logging and Observability page.

The core point the page leads with: the session detail view plays back a **Daily cloud recording of the room the agent ran in**. It is not a separate Pipecat Cloud feature, and Pipecat Cloud adds no recording defaults of its own. Once you know that, the rest follows.

It covers:

- **Requirements**: the Daily transport is needed, since cloud recording is a feature of the Daily room. An agent on a Twilio or generic WebSocket leg has no room to record and gets no player.
- **Turning it on**: `start_cloud_recording` on the meeting token, or imperative `start_recording()` / `stop_recording()`.
- **What the recording contains**: user audio is captured before any in-pipeline audio filter, because the recording is made from what participants publish into the room. An input filter like Krisp VIVA runs inside the pipeline and only cleans the copy feeding STT and VAD.
- **Storage and retention**: recordings are kept until deleted, there is no automatic expiry, and storage is billed per minute per month stored.
- **Custom S3**: `recordings_bucket` with a full example, so lifecycle control sits with the customer's own S3 rules. This is the real answer for a compliance requirement.

## Notes for review

- The retention behavior (kept until deleted, no automatic expiry) is confirmed internally but was not written down anywhere, which is half the reason for this PR. Worth a second pair of eyes.
- The pre-filter claim is verified against `base_input.py`, where `audio_in_filter` is applied to inbound frames inside the input transport, and against the media server, which builds the recording from participant streams.
- All internal links checked against files in this repo. `recordings_bucket` field names checked against `RecordingsBucketConfig` and Daily's REST reference.

Raised by support ticket T-2978.

### DIFF
--- a/pipecat-cloud/fundamentals/session-recordings.mdx
+++ b/pipecat-cloud/fundamentals/session-recordings.mdx
@@ -14,7 +14,7 @@ Two things:
 1. **Your agent runs on the Daily transport.** Cloud recording is a feature of the Daily room your agent is in. An agent on a non-Daily transport (a Twilio or generic WebSocket leg, for example) has no room to record, so no recording is made and no player appears on its session page.
 2. **A cloud recording is actually started for that room.** Pipecat Cloud does not start one on your behalf. See the next section.
 
-Nothing else is required. You do not need to enable the dashboard player separately, and you do not need an `AudioBufferProcessor` in your pipeline (see [Choosing one recording approach](#choosing-one-recording-approach)).
+Nothing else is required. You do not need to enable the dashboard player separately.
 
 ## Turning recording on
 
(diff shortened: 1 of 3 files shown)


pipecat-cloud/fundamentals/session-recordings.mdx
```diff
@@ -0,0 +1,97 @@
+---
```
markbackman, line 1: Pipecat supports many types of recordings, not just Daily. I'm hesitant to include a docs section like this which is just Daily-centric as it excludes other types of recording. For example, if you're using a websocket connection to a telephony provider, this does not apply to you.

Maybe we need a generic section about recording, where one of the options is Daily recordings. We can specify what's needed there.

jamsea: Makes sense, basically I'm trying to explain in writing when/why the recording appears in Pipecat Cloud

PR #1027, opened 2026-07-31
Fix SoundfileMixer docs: it does not resample sound files

jamsea:
## The problem

The SoundfileMixer page lists this under **Notes**:

> Automatically resamples audio files to match output sample rate

That is false, and it has been false since **v0.0.50** (2024-12-11). The changelog for that release says:

> `SoundfileMixer` doesn't resample input files anymore to avoid startup delays. The sample rate of the provided sound files now need to match the sample rate of the output transport.

The current source backs this up. If a sound file's sample rate does not match the output transport, the mixer logs a warning and stores nothing. The mix step then just passes the audio straight through.

## Why it matters

This fails silently. The bot starts fine, joins fine, and talks fine. The only signal is one warning at startup. The customer gets no background audio for the whole session and has no obvious reason why.

A customer hit this on support ticket. They read this page, saw "automatically resamples", and ruled out sample rate as the cause.

Worth noting: the autogenerated reference docs at reference-server.pipecat.ai already say the right thing. Only this Mintlify page is wrong.

## What changed

- Added a `<Warning>` near the top of the page with the real behavior, the actual warning string customers can grep their logs for (`has incorrect sample rate`), and the fix
- Replaced the false "Automatically resamples" bullet in **Notes** with two accurate bullets
- Added the sample rate requirement to the `sound_files` parameter description, since that is where people look when setting it up

I made the warning prominent rather than a passing bullet because the silent failure is the trap here.

### DIFF
--- a/api-reference/server/utilities/audio/soundfile-mixer.mdx
+++ b/api-reference/server/utilities/audio/soundfile-mixer.mdx
@@ -8,24 +8,8 @@ description: "Audio mixer for combining real-time audio with sound files"
 `SoundfileMixer` is an audio mixer that combines incoming audio with audio from files. It supports multiple audio file formats through the soundfile library and can handle runtime volume adjustments and sound switching.
 
 <Warning>
-  **Your sound files must already match the output transport's sample rate. The
-  mixer does not resample them.**
-
-  If a file's sample rate does not match, the mixer skips that file when it loads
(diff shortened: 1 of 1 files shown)


markbackman (approved):
LGTM. Just restructured things and streamlined the text.

PR #1031, opened 2026-07-31
docs(pipecat-cloud): correct what maxSessionDuration actually does (PCC-1066)

mattshep:
Docs half of [PCC-1066](https://linear.app/dailyco/issue/PCC-1066). Companion to [pipecat-cloud-images#102](https://github.com/daily-co/pipecat-cloud-images/pull/102).

## Why

The cap was documented as closing the agent's connection, with the bot "receiving a connection error the next time it tries to send or receive data".

That has never been true for HTTP transports. The platform's `/bot` request is a one-shot POST — the bot never reads or writes on it again, so closing it produces no error the bot can observe and the session simply carries on past its limit. That gap is the bug behind PCC-1066: a customer's session ran ~43 minutes past a 2-hour cap while a second session was placed into the same process.

Base image `0.1.27` enforces the limit inside the bot process, so the outcome the docs already described now actually happens. This PR fixes the four places that described the old mechanism.

## Changes

- **`pipecat-cloud/fundamentals/active-sessions.mdx`** — rewritten to say the bot is cancelled and the pipeline shuts down. Keeps the "not a polite signal / use your own timer" guidance, which is still correct, and adds that the pipeline unwinds cleanly (`cleanup()` runs on each processor) since that is now true and worth knowing. Adds a version note for anyone on an older base image whose sessions currently overrun — they should raise `maxSessionDuration` before upgrading.
- **`api-reference/cli/cloud/deploy.mdx`**, **`openapi.json`** (create + update request schemas) — same correction to the one-line description.
- **`pipecat/learn/pipeline-termination.mdx`** — the cross-reference note.

Left alone: the `maxSessionDuration` **response** field description in `openapi.json`, which describes the configured value rather than the behaviour.

## Sequencing

Safe to merge independently, but the version note names 0.1.27 — ideally land after images#102 is released so the reference is live.

### DIFF
--- a/api-reference/cli/cloud/deploy.mdx
+++ b/api-reference/cli/cloud/deploy.mdx
@@ -82,9 +82,8 @@ In addition to this flag, you also need to enable the `KrispVivaFilter()` for yo
 
 <ParamField path="--max-session-duration" type="number" default="7200">
   Maximum session duration in seconds. When a session reaches this limit, the
-  session is stopped: the bot is cancelled and its pipeline shuts down, with no
-  opportunity for the bot to say goodbye first. Valid range: `60` to `14400` (4
-  hours). Defaults to `7200` (2 hours) when unset.
+  session is stopped: the bot is cancelled, with no opportunity for it to say
+  goodbye first. Valid range: `60` to `14400` (4 hours). Defaults to `7200` (2 hours) when unset.
 
 See [Session duration limits](/pipecat-cloud/fundamentals/active-sessions#session-duration-limits)
 for more information.
@@ -318,8 +317,8 @@ agent_profile = "agent-2x"
 
 <ParamField path="max_session_duration" type="number" default="7200">
 Maximum session duration in seconds. When a session reaches this limit, the
-agent's connection is forcibly closed — the session is cut off mid-flight
-with no notice to the bot. Valid range: `60` to `14400` (4 hours). See
+session is stopped: the bot is cancelled, with no opportunity for it to say
+goodbye first. Valid range: `60` to `14400` (4 hours). See
 [Session duration limits](/pipecat-cloud/fundamentals/active-sessions#session-duration-limits)
 for more information.
 
--- a/api-reference/cli/cloud/deploy.mdx
+++ b/api-reference/cli/cloud/deploy.mdx
@@ -82,9 +82,9 @@ In addition to this flag, you also need to enable the `KrispVivaFilter()` for yo
 
 <ParamField path="--max-session-duration" type="number" default="7200">
   Maximum session duration in seconds. When a session reaches this limit, the
-  agent's connection is forcibly closed — the session is cut off mid-flight
-  with no notice to the bot. Valid range: `60` to `14400` (4 hours). Defaults
-  to `7200` (2 hours) when unset.
+  session is stopped: the bot is cancelled and its pipeline shuts down, with no
+  opportunity for the bot to say goodbye first. Valid range: `60` to `14400` (4
+  hours). Defaults to `7200` (2 hours) when unset.
 
 See [Session duration limits](/pipecat-cloud/fundamentals/active-sessions#session-duration-limits)
 for more information.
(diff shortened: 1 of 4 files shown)


jamsea (approved):
Ah this makes more sense

markbackman (approved):
LGTM

api-reference/cli/cloud/deploy.mdx
cbrianhill, line 329: Should this section be updated for consistency with the above?

pipecat-cloud/fundamentals/active-sessions.mdx
```diff
@@ -333,7 +333,17 @@ Deleting an agent deployment will block any new sessions from starting, but will
...
+
+This is a platform-level safeguard, not a polite signal. The pipeline unwinds cleanly — each processor's `cleanup()` runs — but the bot gets no opportunity to say anything first. If you want it to say goodbye before the cap, implement your own timer in bot code that fires slightly earlier.
+
+<Note>
+  Enforcing the limit inside the bot process requires **base image 0.1.27 or
+  later**. On older base images the limit only closes the platform's connection
+  to your bot, which a bot using an HTTP transport (for example a Daily/WebRTC
+  room) cannot observe — so the session keeps running past its limit. If your
+  agent relies on that, raise `maxSessionDuration` before upgrading.
+</Note>
```
cbrianhill: Should we point out that the invocation of `cleanup()` only applies if the bot uses `PipelineRunner()`? The [comments](https://github.com/daily-co/pipecat-cloud-images/pull/102/changes#diff-821329b7d1de3fd17d9e02adf146a4b27c300aaa6037cad985a623be49d9d6e9R192) in the implementation indicate the behavior isn't unconditional.  But I don't know whether there are customers who avoid our runner classes.

pipecat-cloud/fundamentals/active-sessions.mdx
```diff
@@ -333,7 +333,17 @@ Deleting an agent deployment will block any new sessions from starting, but will
...
+Every service has a maximum session duration. When a session reaches this limit, the session is stopped: your `bot()` function is cancelled and the pipeline shuts down.
+
+This is a platform-level safeguard, not a polite signal. The pipeline unwinds cleanly — each processor's `cleanup()` runs — but the bot gets no opportunity to say anything first. If you want it to say goodbye before the cap, implement your own timer in bot code that fires slightly earlier.
+
+<Note>
+  Enforcing the limit inside the bot process requires **base image 0.1.27 or
+  later**. On older base images the limit only closes the platform's connection
+  to your bot, which a bot using an HTTP transport (for example a Daily/WebRTC
+  room) cannot observe — so the session keeps running past its limit. If your
```
cbrianhill: The term 'HTTP transport' doesn't seem quite correct to me.  Aren't we really talking about HTTP-activated bots?  (i.e. an HTTP request starts the session, and then the bot configures its own transports for media)

cbrianhill (approved):
A couple of minor comments for clarity, but this overall looks great.

mattshep:
All three addressed.

**The `pcc-deploy.toml` field** — you're right, I'd missed it. My grep matched the CLI flag section and stopped there; the TOML `ParamField` further down had the same stale description. Fixed.

**`cleanup()` is conditional** — correct, and worth being precise about rather than dropping. The implementation cancels `bot()`; whether the pipeline unwinds cleanly depends on `PipelineRunner` absorbing the cancellation. A bot that doesn't use it just sees a `CancelledError` wherever it was awaiting. Now says both, since the second case is the one where a reader needs to do something. (To your question: I don't know either. Worth telling people what happens in both cases regardless.)

**"HTTP transport"** — agreed, and it's a better catch than a wording nit. The HTTP request only *starts* the session; the bot then connects its own media transport. That distinction is precisely why the old behaviour was broken — we were closing the activation request and expecting a bot to notice, when the thing carrying its audio was somewhere else entirely. Reworded here, and in the base-image code comments (daily-co/pipecat-cloud-images#102), which had inherited the same sloppiness from me.

Also normalised the one-line descriptions in `openapi.json` and `deploy.mdx` so they don't promise the pipeline shutdown unconditionally either.

PR #1049, opened 2026-08-03
docs: add a memory troubleshooting guide, and clarify that sequential sessions reuse the same process (T-2660)

jamsea:
Found while working Daily support ticket **T-2660**: a customer saw memory grow call after call on a warm Pipecat Cloud instance.

## The gap

Nothing public said that a warm instance which picks up a new session is running the **same process** that handled the previous session. Worse, the Session API page said the opposite:

> Different sessions are separate pods and don't share state.

That sentence is about routing, and it is true for two sessions running at the same time. But it reads as a promise of a fresh process per session, which is not what happens for sequential sessions on a reused warm instance. The scaling page says an instance "is returned to the pool and can immediately serve another session" and never says the process survives.

And nothing told a developer what to do about it. There was no page on memory at all.

## What changed

**`pipecat/deployment/troubleshooting-memory.mdx`** (new)

A short troubleshooting page. It splits memory problems into the two shapes they actually take, because the fixes are different:

- **Growth inside one call.** `AudioBufferProcessor` with the default `buffer_size=0` holds the whole recording in memory until recording stops, and it pads each track with silence to keep the two tracks aligned, so the buffers grow with the length of the call rather than with how much anyone said. At 24 kHz 16-bit that is ~2.9 MB per minute per track, so a 30-minute call is ~170 MB against a 1 GB `agent-1x` profile. Also: an LLM context that never shrinks, and lists you append to every turn.
- **Growth across calls.** Per-session state left on module scope, clients never closed, tasks never cancelled, handlers never removed.

Then a short "How to measure" section: per-session CPU and memory metrics, a `tracemalloc` snapshot across one long session, and several sessions back to back against the same process.

Registered in `docs.json` under Pipecat → Deployment.

**`pipecat-cloud/guides/session-api.mdx`**

Split the "Session scope" bullet in two. The first keeps the real guarantee: two sessions running at the same time are on separate instances, so a request for one never reaches the other. The second says plainly that being routed separately is not the same as getting a fresh process, and links to the new scaling section.

**`pipecat-cloud/fundamentals/scaling.mdx`**

New subsection under "Updating scaling configuration": a `min-agents` or `max-agents` change only resizes the pool, an instance that keeps running keeps its process, and what to release when a session ends. Links out to the new memory page for the full checklist, to the merged #1080 section for the no-op deploy and `--force` detail, and to `readyz()` for taking one instance out of rotation (which does not restart the process).

## Verified against source

Checked on current `main` of each repo rather than from memory:

- The base image runs one long-lived server and dispatches each session into that same process as a background task. Nothing exits or restarts the process when a session ends, and module-level state persists.
- A deploy that changes only the autoscaling values takes the version-only path: no new ReplicaSet, no pod replacement.
- `AudioBufferProcessor` in `src/pipecat/processors/audio/audio_buffer_processor.py`: `buffer_size=0` means the flush branch never runs, so `_user_audio_buffer` and `_bot_audio_buffer` accumulate until `stop_recording()`; `_sync_buffer_to_position` and `_fill_buffer_silence_gap` pad silence so both buffers track wall-clock, not speech.
- Evidence from the customer's namespace: one instance started once, never restarted its container, and served 7 sessions across roughly 2.5 days.

## Overlap, resolved

Rebased onto current `main`. #1028, which covered the same finding from the `readyz()` angle, is now closed, so there is nothing left to dedupe there. #1080 has since merged the no-op deploy and `--force` explanation into `deploy.mdx`, so this branch drops its own copy of that and links to it instead.

`mint broken-links` clean, `docs-meta-lint` 0 errors, `llms.txt` and `llms-full.txt` regenerated.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -163,6 +163,7 @@
                 ]
               },
               "pipecat/deployment/telephony-in-production",
+              "pipecat/deployment/troubleshooting-memory",
               {
                 "group": "Hosting Platforms",
                 "pages": [
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -11725,7 +11725,7 @@ Roark connects to Pipecat through a single observer, then layers testing and mon
     ```python
     import os
 
-    from pipecat.pipeline.task import PipelineParams, PipelineTask
+    from pipecat.pipeline.worker import PipelineWorker
     from pipecat_roark import RoarkObserver
 
     roark = RoarkObserver(
@@ -11742,7 +11742,7 @@ Roark connects to Pipecat through a single observer, then layers testing and mon
         context_aggregator.assistant(),
     ])
 
-    task = PipelineTask(pipeline, params=PipelineParams(observers=[roark]))
+    worker = PipelineWorker(pipeline, observers=[roark])
     ```
 
     <Tip>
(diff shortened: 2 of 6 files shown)


markbackman:
I'm inclined to suggest that we close this.

PR #1050, opened 2026-08-03
docs: webhook URL takes the organization name, not an org id

jamsea:
## The bug

The Pipecat Cloud "Daily PSTN Dial-in" guide documented the pinless `room_creation_api` webhook URL like this:

```
https://api.pipecat.daily.co/v1/public/webhooks/{org_id}/{agent_name}/dialin
```

That first path segment is not an id. The API resolves the organization by **name**. The route is declared with an `:organizationName` param and the handler looks the org up by its name, so a UUID never matches.

## Customer impact

Someone reads `{org_id}`, pastes their organization UUID, and every dial-in call fails. The webhook returns `404 "Service not found"` and there is nothing useful in the logs to point at the URL. A real customer hit this (T-3093).

## The fix

Placeholders are now `ORGANIZATION_NAME` / `AGENT_NAME`, with a short line saying where the name comes from: `pipecat cloud organizations list`. That matches the `AGENT_NAME.ORGANIZATION_NAME` wording the other telephony pages already use for `serviceHost`.

## Files changed

- `pipecat-cloud/guides/telephony/daily-dial-in.mdx`: the reported bug.
- `pipecat-cloud/guides/whatsapp.mdx`: same URL, same route, was using `$ORGANIZATION_ID`.
- `pipecat/features/whatsapp.mdx`: same as above.

The two WhatsApp pages hit the exact same `/v1/public/webhooks/:organizationName/:serviceName/...` route, so they had the same bug. Fixed them here rather than leaving the copy-pasted version wrong.

### DIFF
--- a/pipecat-cloud/guides/telephony/daily-dial-in.mdx
+++ b/pipecat-cloud/guides/telephony/daily-dial-in.mdx
@@ -20,6 +20,8 @@ Let Pipecat Cloud handle the webhook automatically.
 
 **Using the REST API:**
 
+Get your organization name with `pipecat cloud organizations list`.
+
 ```bash
 curl --location 'https://api.daily.co/v1' \
 --header 'Authorization: Bearer YOUR_DAILY_API_KEY' \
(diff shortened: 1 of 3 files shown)


markbackman:
Correcting in the CLI instead, to match the actual names:
https://github.com/daily-co/pipecat-cloud/pull/189

Closing.

PR #1057, opened 2026-08-05
docs: qualify the 10 second cold start number

jamsea:
## Problem

The docs state a flat "around 10 seconds" for a Pipecat Cloud cold start. A repo-wide grep found the claim in exactly three places, all in `pipecat-cloud/`:

1. `pipecat-cloud/fundamentals/scaling.mdx` (the `#cold-starts` accordion, "How long does a cold start take?")
2. `pipecat-cloud/fundamentals/scaling.mdx` (the `#controlling-costs` section, "a ~10-second cold start")
3. `pipecat-cloud/guides/capacity-planning.mdx` (a bare `<Info>` line under cost-efficient scaling strategies)

Found while working support ticket T-3095. A customer running a 1.5 GB image measured a 65 second cold start and had planned their user experience around the 10 second number. The scaling page does mention elsewhere that duration "depends on multiple factors", but the headline number is what people plan against, and one of the three spots had no qualifier at all.

## Changes

* 10 seconds is now framed as a best case (small image, agent that starts up fast) rather than a typical value, and the docs say plainly that image size and agent setup work can push it to a minute or more.
* New accordion on the scaling page titled "My cold start takes much longer than 10 seconds. Why?" so the page is findable by the symptom. It gives four things a customer can act on:
  * Trim the image, use multi-stage builds.
  * Everything at the module level in `bot.py` runs before the instance can accept a session, so an import that loads a model, or a model created outside `bot()`, is part of the cold start. It is not only the code inside `bot()`.
  * Bake model and VAD weights into the image at build time instead of fetching them on the first call.
  * Keep a warm instance with `min-agents` and plan a waiting experience for the times a cold start still happens.
* The capacity planning callout now links to the cold starts section instead of dead-ending on a bare number.

## Deliberately not done

No replacement range like "30 to 50 seconds" is published. There is no instrumented measurement or SLO for bot start time to back a new authoritative number, so this PR qualifies the existing one and names the factors instead of inventing a spec.

No customer or ticket is named in the doc text.

## Notes

Prettier run on both files. Draft PR: not ready for review.

### DIFF
--- a/pipecat-cloud/fundamentals/scaling.mdx
+++ b/pipecat-cloud/fundamentals/scaling.mdx
@@ -144,9 +144,25 @@ Pipecat Cloud aims to mitigate cold starts as much as possible through [auto-sca
 
 <AccordionGroup>
 <Accordion title="How long does a cold start take?">
-The exact duration of a cold start depends on multiple factors, such as the size of your agent image and the complexity of your agent. In general, cold starts take around 10 seconds. Once you have a running instance of a deployed image, new sessions start immediately with no cold start delay.
+Around 10 seconds is a best case: a small agent image with an agent that finishes its setup work fast. Treat it as a floor, not a promise.
+
+Two things move the number a lot:
+
+- **Image size.** The image has to be pulled and started before your agent runs. A large image can push a cold start to a minute or more.
+- **Startup work in your agent.** Anything your agent does before it can take a session adds to the wait. Loading models, downloading weights, and opening connections all count.
+
+Once you have a running instance of a deployed image, new sessions start immediately with no cold start delay.
 
 For developers, especially during test and development, using scale-to-zero with cold starts is a good place to start as it saves money. Once you understand your utilization pattern, you can build a strategy around optimizing start times by maintaining warm agent instances.
+
+</Accordion>
+<Accordion title="My cold start takes much longer than 10 seconds. Why?">
+A cold start that runs far longer than 10 seconds, say 30 seconds or a minute or more, usually comes down to your image or your agent's startup code. Things to check:
+
+- **How big is your image?** Every megabyte has to be pulled before your agent can run. Drop what the bot does not need at runtime, and use multi-stage builds to keep build tools out of the final image.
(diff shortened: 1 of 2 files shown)


pipecat-cloud/fundamentals/scaling.mdx
```diff
@@ -144,11 +144,20 @@ Pipecat Cloud aims to mitigate cold starts as much as possible through [auto-sca
...
 
-For developers, especially during test and development, using scale-to-zero with cold starts is a good place to start as it saves money. Once you understand your utilization pattern, you can build a strategy around optimizing start times by maintaining warm agent instances.
+A cold start that runs far longer than 10 seconds, say 30 seconds or a minute or more, usually comes down to your image or your agent's startup code. Things to check:
+
+- **How big is your image?** Every megabyte has to be pulled before your agent can run. Drop what the bot does not need at runtime, and use multi-stage builds to keep build tools out of the final image.
+- **What runs at import time in `bot.py`?** Everything at the module level runs before the instance can accept a session. So an import that loads a model, or a model you create outside your `bot()` function, is part of the cold start. It is not only the code inside `bot()` that costs you time.
+- **Are you fetching model files at runtime?** Bake model and VAD weights into the image at build time. Then the instance starts with the files already on disk instead of downloading them on the first call.
```
markbackman, line 153: You shouldn't be running models in Pipecat, usually. I don't think we want to give that impression.

markbackman (approved):
LGTM. Just one comment.

pipecat-cloud/fundamentals/scaling.mdx
jamsea: @markbackman IIRC doesn't stuff like Silero download at run time if it's not detected or have we changed how that works?

markbackman: Yes, but that download time is small, a few MBs, IIRC.

The paragraph reads as if you can run models on PCC, which is not the case. I'd like to avoid potential confusion. If you want an LLM model for inference, you should be running that on a dedicated endpoint.

PR #1062, opened 2026-08-07
Add Gandr TTS as a community integration

Alg1312:
Adds **Gandr** to the TTS table as a community integration, with a service reference page and the navigation entry.

### What it is

A text-to-speech service. The integration subclasses `WebsocketTTSService` and holds one connection open across turns rather than opening one per utterance. The first turn on a fresh connection runs about 700 ms while the session voice cache fills. Every turn after it is materially quicker.

- **Repo:** https://github.com/Gandr-AI/gandr-pipecat
- **License:** MIT
- **Install:** `uv pip install git+https://github.com/Gandr-AI/gandr-pipecat.git`

### Against the community-integration requirements

| Requirement | Status |
|---|---|
| Source following Pipecat patterns | `GandrTTSService(WebsocketTTSService)` |
| Foundational example | `examples/foundational/gandr_tts_basic.py` |
| README with install/usage/compatibility | yes |
| Permissive license | MIT |
| Docstrings | every public definition in `tts.py` |
| Changelog | `CHANGELOG.md` |

### Verification

**Tested with Pipecat v1.7.0** on Python 3.12.11. Clean virtual environment: `pip install` exit 0, imports, **11/11 unit tests pass**, service constructs.

Worth noting for reviewers: the package declares `pipecat-ai>=0.0.108` and was written against that version, so this test crosses a major version and about a year of API drift. `WebsocketTTSService` is still the correct base class and the interface still fits.

**Not claimed:** no live socket was opened in that run. It covers install, import, unit tests and construction. Rendering audio costs money, so it is left to whoever runs it.

### Behaviour a reviewer may want to know about

- **Barge-in drops and reopens the connection.** The wire protocol has no cancel frame, so the alternative was draining audio nobody will hear. Configurable via `reconnect_on_interruption=False`.
- **Every failure path closes the turn with `TTSStoppedFrame`**, so a pipeline cannot hang on a dead connection.
- **The 2,000-character request cap is handled by a lossless splitter**, so callers do not pre-chunk. Its 11 tests were proved red under 8 mutations. Two survived their first mutant and were rewritten, having been green for the wrong reason.
- **`language` takes bare two-letter codes.** A region suffix such as `en-GB` is not recognised and falls back to English without erroring, so the reference page carries an explicit warning rather than leaving it to be discovered.

### Demo video

Not included in this PR. Happy to add one if it is a blocker for merge. Flag it and I will attach a 30-60s clip.

### DIFF
--- a/api-reference/server/services/supported-services.mdx
+++ b/api-reference/server/services/supported-services.mdx
@@ -197,7 +197,7 @@ Text-to-Speech services receive text input and output audio streams or chunks.
 | [ElevenLabs](/api-reference/server/services/tts/elevenlabs)       | `uv add "pipecat-ai[elevenlabs]"`                                             | Pipecat    |
 | [Fish](/api-reference/server/services/tts/fish)                   | `uv add "pipecat-ai[fish]"`                                                   | Pipecat    |
 | [Floe](/api-reference/server/services/tts/floe)                   | `uv add pipecat-floe`                                                         | Community  |
-| [Gandr](/api-reference/server/services/tts/gandr)                 | `uv pip install git+https://github.com/Gandr-AI/gandr-pipecat.git`            | Community  |
+| [Gandr](/api-reference/server/services/tts/gandr)                 | `uv add pipecat-gandr`                                                        | Community  |
 | [Gnani](/api-reference/server/services/tts/gnani)                 | `uv add pipecat-gnani`                                                        | Community  |
 | [Google](/api-reference/server/services/tts/google)               | `uv add "pipecat-ai[google]"`                                                 | Pipecat    |
 | [Gradium](/api-reference/server/services/tts/gradium)             | `uv add "pipecat-ai[gradium]"`                                                | Pipecat    |
--- a/api-reference/server/services/supported-services.mdx
+++ b/api-reference/server/services/supported-services.mdx
@@ -197,6 +197,7 @@ Text-to-Speech services receive text input and output audio streams or chunks.
 | [ElevenLabs](/api-reference/server/services/tts/elevenlabs)       | `uv add "pipecat-ai[elevenlabs]"`                                             | Pipecat    |
 | [Fish](/api-reference/server/services/tts/fish)                   | `uv add "pipecat-ai[fish]"`                                                   | Pipecat    |
 | [Floe](/api-reference/server/services/tts/floe)                   | `uv add pipecat-floe`                                                         | Community  |
+| [Gandr](/api-reference/server/services/tts/gandr)                 | `uv pip install git+https://github.com/Gandr-AI/gandr-pipecat.git`            | Community  |
 | [Gnani](/api-reference/server/services/tts/gnani)                 | `uv add pipecat-gnani`                                                        | Community  |
 | [Google](/api-reference/server/services/tts/google)               | `uv add "pipecat-ai[google]"`                                                 | Pipecat    |
 | [Gradium](/api-reference/server/services/tts/gradium)             | `uv add "pipecat-ai[gradium]"`                                                | Pipecat    |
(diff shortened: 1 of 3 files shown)


api-reference/server/services/tts/gandr.mdx
```diff
@@ -0,0 +1,180 @@
...
+  </Card>
+</CardGroup>
+
+## Installation
+
+This is a community-maintained package distributed separately from
+`pipecat-ai`:
+
+```bash
+uv pip install git+https://github.com/Gandr-AI/gandr-pipecat.git
```
markbackman: Looks like you have a pypi package, so this should be:
```
uv add pipecat-gandr
```

api-reference/server/services/supported-services.mdx
```diff
@@ -197,6 +197,7 @@ Text-to-Speech services receive text input and output audio streams or chunks.
 | [ElevenLabs](/api-reference/server/services/tts/elevenlabs)       | `uv add "pipecat-ai[elevenlabs]"`                                             | Pipecat    |
 | [Fish](/api-reference/server/services/tts/fish)                   | `uv add "pipecat-ai[fish]"`                                                   | Pipecat    |
 | [Floe](/api-reference/server/services/tts/floe)                   | `uv add pipecat-floe`                                                         | Community  |
+| [Gandr](/api-reference/server/services/tts/gandr)                 | `uv pip install git+https://github.com/Gandr-AI/gandr-pipecat.git`            | Community  |
```
markbackman: Same here:
```suggestion
| [Gandr](/api-reference/server/services/tts/gandr)                 | `uv add pipecat-gandr`            | Community  |
```

markbackman (commented):
LGTM. Just two minor changes, then this will be ready.

api-reference/server/services/tts/gandr.mdx
markbackman: Wait, maybe you don't have a PyPI package. That means that your repo is inaccurate. Either way, please review to make sure the right information is in both places, so developers have an easy time figuring out how to use your integration.

Alg1312:
Thanks, you are right on both counts, and the second comment is the useful one.

There is no PyPI package yet, and the repo was the thing that was wrong: its README said pip install pipecat-gandr, which fails. Anyone following it hit an error on the first command. Fixed in Gandr-AI/gandr-pipecat@6592803. The README now gives the git install, so it matches this PR and actually works.

I left the install line here as git+https://... rather than applying your suggestion, because uv add pipecat-gandr would resolve to nothing today. That is the same inaccuracy pointed the other way. A couple of the existing community entries (Uplift AI, Voice.ai) use the git form, so I do not think it blocks listing.

We are publishing to PyPI. The name is unclaimed and the package is hatchling-built and ready. It is waiting on an account credential on our side, not on any work. The moment it resolves I will push a commit here changing both lines to uv add pipecat-gandr and update the repo README in the same change.

Happy to do it either way. If you would rather hold the PR until that lands, say so and I will leave it open until the package is live.

markbackman:
These docs won't go live until the next Pipecat release, so you should have time to publish and update these docs. Please let me know when it's ready.

Alg1312:
Package is live: https://pypi.org/project/pipecat-gandr/ (0.1.0).

Both install lines in this PR now say uv add pipecat-gandr: the install block in gandr.mdx and the table row in supported-services.mdx. The repo README got the same change at the same time, so the two match.

Verified from a clean venv: pip install pipecat-gandr from PyPI, import, exit 0, version 0.1.0.

And thanks for pushing on the README. It had been telling people to run an install that did not resolve, which is the worst command in the docs to get wrong.

markbackman (approved):
Nice! Thanks for the quick turn around. Approving.

PR #1071, opened 2026-08-08
Gandr TTS: stable default URL, language count, copy fixes

Alg1312:
Follow-up to the merged Gandr TTS page, three small fixes before it goes live with the next release.

1. The default url now points at the stable hostname (wss://tts.gandr.ai/ws), matching the OpenAPI contract and the package default. Both names resolve to the same service today, so this is a naming fix, not a behavior change.
2. Language support was under-stated: 23 languages are supported, and the listed codes are examples rather than the full set.
3. A few dashes replaced with plain punctuation for house style.


markbackman (approved):
👍

PR #1076, opened 2026-08-10
Fix observer examples: observers goes on PipelineWorker, not PipelineParams

jamsea:
## What is wrong

Several observer pages showed `observers=[...]` passed inside `PipelineParams(...)`:

```python
worker = PipelineWorker(
    pipeline,
    params=PipelineParams(
        observers=[DebugLogObserver()],
    ),
)
```

`observers` is not a field of `PipelineParams`. It is an argument on `PipelineWorker` (and on the older `PipelineTask`, which subclasses it).

## What changed

Moved `observers` out to the worker or task call on all 11 places that had it wrong, across 7 pages:

- `api-reference/server/utilities/observers/debug-observer.mdx` (3)
- `api-reference/server/utilities/observers/user-bot-latency-observer.mdx` (2)
- `pipecat/evals/platforms/roark.mdx` (2)
- `api-reference/server/utilities/observers/llm-observer.mdx` (1)
- `api-reference/server/utilities/observers/transcription-observer.mdx` (1)
- `api-reference/server/utilities/observers/observer-pattern.mdx` (1)
- `api-reference/server/services/analytics/roark.mdx` (1)

## How this was checked

Grepped every `.md` and `.mdx` file in the repo for `observers=`. 19 hits total: 11 were wrong and are fixed here, 8 were already correct and are untouched (`pipeline-params.mdx`, `startup-timing-observer.mdx`, `turn-tracking-observer.mdx`, `pipecat/fundamentals/metrics.mdx`, and two prose mentions).

Verified against pipecat `main` at commit `a20f386d9`:

- `src/pipecat/pipeline/worker.py:248` declares `observers` in `PipelineWorker.__init__`, used at `:352` and `:485`.
- `src/pipecat/pipeline/worker.py:137` defines `PipelineParams`, which has no `observers` field. Its own docstring says to use `PipelineWorker` constructor arguments for other worker parameters.
- `examples/observability/observability-observer.py:142-160` shows the correct shape: `observers=[...]` alongside `params=PipelineParams(...)`.
- `class BaseOutputTransport` lives in `src/pipecat/transports/base_output.py:60`. There is no `base_output_transport` module.

## Not in scope

`PipelineTask` is deprecated since 1.3.0 in favor of `PipelineWorker`. The Roark pages still use `PipelineTask`. Swapping those over is a separate change, so this PR leaves them alone and only fixes where `observers` is passed.

### DIFF
--- a/api-reference/server/services/analytics/finchvox.mdx
+++ b/api-reference/server/services/analytics/finchvox.mdx
@@ -65,7 +65,7 @@ self-hosted option.
 
 Enabling observability in a Pipecat app involves three steps: initialize
 Finchvox, add the `FinchvoxProcessor` to your pipeline, and enable metrics,
-tracing, and turn tracking on your `PipelineTask`.
+tracing, and turn tracking on your `PipelineWorker`.
 
 1. Initialize Finchvox at the top of your bot (e.g., `bot.py`):
 
@@ -88,11 +88,11 @@ pipeline = Pipeline([
 ])
 ```
 
-3. Initialize your `PipelineTask` with metrics, tracing, and turn tracking
+3. Initialize your `PipelineWorker` with metrics, tracing, and turn tracking
    enabled:
 
 ```python
-task = PipelineTask(
+worker = PipelineWorker(
     pipeline,
     params=PipelineParams(enable_metrics=True),
     enable_tracing=True,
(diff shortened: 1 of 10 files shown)


markbackman (approved):
LGTM! Thank you.

PR #1079, opened 2026-08-11
docs(rime): migrate Arcana examples to Coda

naszzz:
## Summary
- remove current Arcana guidance from the Rime TTS reference
- document the Coda sampling settings
- use Coda with Luna in the Rime examples because Pipecat already used Luna and Rime supports it on both Arcana and Coda
- update the matching llms-full.txt content


markbackman (approved):
LGTM. Thanks for the updates 🙇

PR #1080, opened 2026-08-11
Document when a deploy does not replace running agents

jamsea:
## Why

Nothing customer-facing documented the fact that a Pipecat Cloud deploy does not always replace running pods. A customer with warm instances (`min-agents >= 1`) can keep serving an older version of their agent with no sign that anything is wrong. One customer ran a broken Krisp setup for two months, building from the mutable `dailyco/pipecat-base:latest` tag.

The behavior is intentional (a no-op deploy used to roll pods and drop in-flight sessions), but it was only intuitive if you already knew the internals. This documents the customer-facing shape of it.

### DIFF
--- a/pipecat-cloud/fundamentals/deploy.mdx
+++ b/pipecat-cloud/fundamentals/deploy.mdx
@@ -149,46 +149,23 @@ remains consistent and avoids potential cold starts.
 
 ### When a deploy does not replace running agents
 
-A deploy only replaces your running agent instances when it creates a new version of your agent. If the configuration you send is the same as what the agent already has, Pipecat Cloud keeps the current version and leaves your instances alone. That is on purpose: a routine deploy can't cut off a session that is already in progress.
+A deploy only replaces your running agent instances when it creates a new version of your agent. If the configuration you send matches what the agent already has, Pipecat Cloud keeps the current version and leaves your instances alone, so a routine deploy can't cut off a session that is already in progress.
 
-Watch for this if you keep warm instances (`min-agents` set to 1 or higher). Those instances can keep serving your old code, and the deploy output won't look wrong.
+Watch for this if you keep warm instances (`min-agents` set to 1 or higher): they can keep serving your old code, your deploy will report success, and the output won't look wrong.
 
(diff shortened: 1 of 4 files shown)


pipecat-cloud/fundamentals/scaling.mdx
```diff
@@ -241,6 +241,8 @@ pipecat cloud deploy [agent-name] [image] --min-agents 1 --max-agents 5
 Please note that changing your scaling parameters will not disrupt any active sessions.
 If you reduce your max instance count below the number of currently active sessions, you will still be billed for the duration of those sessions.
 
+A scaling-only change also doesn't create a new version or restart your instances, so it won't pick up new code, a new image, or a new secret value ([why](/pipecat-cloud/fundamentals/deploy#when-a-deploy-does-not-replace-running-agents)).
```
markbackman: Has this been a point of confusion? New code and images are only used when they're pushed. This seems obvious, so I want to make sure this has been confusing.

pipecat-cloud/fundamentals/deploy.mdx
```diff
@@ -147,6 +147,26 @@ Idle agent instances in your agent pool will be replaced with the new configurat
 New agent requests may, therefore, start with prior deployment configuration if updates are not fully propagated. This ensures on-demand availability
 remains consistent and avoids potential cold starts.
 
+### When a deploy does not replace running agents
```
markbackman, line 150: @jamsea can we consolidate this point to 1 simple paragraph?

pipecat-cloud/fundamentals/scaling.mdx
jamsea: @markbackman it's the "new secret value" that confuses people (literally just responded to a person asking this a few minutes ago) 

I think we need to remove this line from these docs though and update Pipecat Cloud Dashboard with better copy. Right now when you set a new secret in the Pipecat Cloud Dashboard UI, it tells you to redeploy. If you click the redeploy button on the agent, it doesn't actually update the secret.

PR #1089, opened 2026-08-13
docs(pipecat-cloud): document that min-agents changes apply asynchronously (T-3351)

jamsea:
## Why

A customer raised `min-agents` on their agent, waited 5 seconds, then started dialing. They got repeated 18 to 26 second cold starts and could not work out why, because their scaling config looked right.

The reason: a `min-agents` change is applied asynchronously. The successful response means the change was accepted, not that the new reserved instances exist or are ready to take a session. There is no readiness gate on that response. Their calls raced the scale-up.

Nothing in the public docs said this, and nothing told a customer how to check when the new reserved instances are actually ready. The ~30 second warm-up delay was already documented for idle agents on the capacity planning page, but it was never connected to a `min-agents` change.

## Searchability

The new section names the symptom a customer actually sees ("unexpected cold starts right after raising min-agents") so it turns up in search and in Kapa answers.


markbackman:
I think this point is already pretty clearly spelled out in other places. I'm going to close this.

PR #1107, opened 2026-08-19
docs: fix service import paths that raise AttributeError (T-2331)

jamsea:
## The symptom a customer hits

Copy the import off a service page and run it:

```python
from pipecat.services.nvidia import NvidiaLLMService
```

```
AttributeError: module 'pipecat.services.nvidia' has no attribute 'NvidiaLLMService'
```

## Why

Flat service imports were removed in Pipecat 1.0 (documented in the 1.0 migration guide). None of the vendor packages under `pipecat.services` re-export their classes, so the short form always fails. The working form is the submodule path, which is also what the CLI registry (`src/pipecat/cli/registry/_imports.py`) uses as canonical:

```python
from pipecat.services.nvidia.llm import NvidiaLLMService
```

## What changed

124 import lines across 62 service reference pages and 1 fundamentals page, plus the matching 124 lines in `llms-full.txt`.

Left alone on purpose:

- The `# Before (1.0)` example in `pipecat/migration/migration-1.0.mdx`. It is supposed to show the old form.
- Module level imports that already work and were verified: `pipecat.services.llm_service`, `pipecat.services.stt_service`, `pipecat.services.tts_service`, `pipecat.services.mcp_service`.

## Verification

Pipecat repo on `main` at `1.7.1.dev562`, Python 3.12.9, using that repo's `.venv`. Every short form raised `AttributeError`. Every long form passed.

### DIFF
--- a/api-reference/server/services/image-generation/fal.mdx
+++ b/api-reference/server/services/image-generation/fal.mdx
@@ -107,7 +107,7 @@ Runtime-configurable settings passed via the `settings` constructor argument usi
 
 ```python
 import aiohttp
-from pipecat.services.fal import FalImageGenService
+from pipecat.services.fal.image import FalImageGenService
 
 async with aiohttp.ClientSession() as session:
     image_gen = FalImageGenService(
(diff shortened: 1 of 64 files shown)


markbackman (approved):
Nice find!

PR #1118, opened 2026-08-20
docs: fix dial-out request form link and scope approval to phone numbers

DominicStewart:
## What

Daily gates dial-out approval on the **destination**, not the provider. Dialing a phone number needs approval on the domain; dialing a `sip:` URI never has. Our telephony docs did not say this, and the PSTN page pointed at the wrong form.

- **`pipecat/telephony/daily-pstn.mdx`**: the dial-out request link was a Google Forms **`/edit`** URL (`.../1EHy2wsX.../edit`), which redirects to `viewform?edit_requested=true`, so readers were being asked to request edit access. It is the right form, but the wrong URL for it. Now points at the form's public `viewform` URL, and scopes the requirement to phone numbers.
- **`pipecat/telephony/daily-sip.mdx`**: adds a note that SIP URI dial-out needs no approval.
- **`pipecat/telephony/twilio-daily-sip.mdx`**: same note in the dial-out section, since which rule applies depends on the destination you pass to `start_dialout`, not on Twilio.
- **`pipecat-cloud/guides/telephony/daily-dial-out.mdx`**: adds the `allow_dialout` error to the troubleshooting table. It was the one dial-out error the table did not cover, and Pipecat Cloud orgs are created with the flag off by default.

## Why

In `pluot-core`, `skyliner-express/app/sfu/Room.js` checks the two flags differently:

```js
// line 1786: domain flag, only when the destination is a phone number
if (isSipPstnDialOut && !this.domainProps.allow_dialout && phoneOrSipUriDialOut.startsWith('+')) { ... }

// line 1797: room flag, for all dial-out including sip:
if (isSipPstnDialOut && !this.roomProps.enable_dialout) { ... }
```

The same destination-based rule applies to call transfer at line 7898.

Confirmed by Rajneesh and Varun in [#projects-pstn-sip](https://pluot.slack.com/archives/C0492TDCLAY/p1787238124895649): "dialout to `sip:` is default enabled for all", and "It's based on destination address provided in the start_dialout. Nothing to do with what provider is used." Varun noted the change shipped earlier this year and the docs were missed.

Pipecat Cloud orgs get `allow_dialout = '0'` at creation (`skyliner-express/app/services/cloud/domainManager.ts:402`), and the flag is not visible via the API, so a PCC customer doing SIP dial-out sees a gate in the dashboard that does not actually apply to them.

## Impact

Nervecentre submitted the dial-out request form three times, twice for SIP, which they did not need.

The matching Daily Video docs change is in `daily-co/pluot-core`.

Linear: [SOL-251](https://linear.app/dailyco/issue/SOL-251/docs-dial-out-approval-is-documented-as-required-for-all-dial-out-but)
Daily Video counterpart: daily-co/pluot-core#14379


markbackman (approved):

pipecat/telephony/daily-pstn.mdx
```diff
@@ -223,9 +223,11 @@ async def bot(runner_args: RunnerArguments):
 Dial-out allows your bot to initiate calls to phone numbers. Unlike dial-in, your bot starts the call rather than waiting for incoming calls.
 
 <Warning>
-  You must contact Daily to enable dial-out for your account. Submit a support
-  request
-  [here](https://docs.google.com/forms/d/1EHy2wsX20a2HjljAsJkyPqvATIM4WCvm7JzYKypGYis/edit).
+  Dial-out to a phone number requires approval on your Daily domain. Fill out the
+  [dial-out request form](https://docs.google.com/forms/d/e/1FAIpQLSe3hTnIchbS1NgOsO8HGrzyBk5B1WuwWT_qqfNQde14_L-Y_Q/viewform) to request it.
+
+  Dial-out to a `sip:` URI does not require this approval. See
```
vr000m, line 229: I think it's sip_uri. And for phone numbers the api expects that you've been approved to dial-out or it will fail.

markbackman: I'll fix this in a follow up.

PR #1128, opened 2026-08-24
Add missing transports to the Transports page

aconchillo:
## Why

The card grid on `pipecat/learn/transports.mdx` listed 7 transports while the supported services table listed 11, so Beyond Presence, LemonSlice, Vonage Video Connector, and WhatsApp had no card.

## What

- **Supported services table**: list transports by service name instead of Python class name (`DailyTransport` → `Daily`), matching every other table on that page.
- **Transports page**: add the 4 missing cards in the table's alphabetical order, and align the existing titles to the same service names. Also fixes `Websocket` → `WebSocket` casing to match the target pages' `sidebarTitle`s.
- **llms.txt / llms-full.txt**: regenerated. This also picks up the Vakyam TTS entry, which was already missing from the checked-in `llms.txt` before this branch.

Naming follows the official product names: `Beyond Presence` and `Vonage Video Connector`.

## Checks

`mint broken-links` and `scripts/docs-meta-lint.mjs` both pass (0 errors; the 183 warnings are pre-existing).

### DIFF
--- a/api-reference/server/services/supported-services.mdx
+++ b/api-reference/server/services/supported-services.mdx
@@ -258,19 +258,19 @@ Translation services enable real-time speech-to-speech and speech-to-text transl
 
 Transports exchange audio and video streams between the user and bot.
 
-| Service                                                                                 | Setup                                         | Maintainer |
-| --------------------------------------------------------------------------------------- | --------------------------------------------- | ---------- |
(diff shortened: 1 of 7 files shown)


markbackman (approved):
Ship it!

PR #1130, opened 2026-08-24
Move Pipecat Flows into the Pipecat tab

aconchillo:
## Why

Pipecat Flows had a top-level tab alongside Pipecat, Clients, and Cloud, which presented it as a separate product. It's a Pipecat module (`pipecat.flows`), so it should have a group inside the Pipecat tab the way Evals does.

## What

**Content moves to `pipecat/flows/`**, mirroring the `pipecat.flows` module and the flat `pipecat/evals/` layout. The `guides/` subfolder is dropped since the module is small.

| Before | After |
|---|---|
| `pipecat-flows/introduction` | `pipecat/flows/introduction` |
| `pipecat-flows/guides/*` (5) | `pipecat/flows/*` |
| `pipecat-flows/examples` | `pipecat/flows/examples` |
| `pipecat-flows/migration/migration-1.0` | `pipecat/migration/flows` |

**Nav** is `Flows`, matching the `pipecat.flows` module, the `flows.pipecat.ai` editor, and the existing `Pipecat Flows` group in the API Reference tab — plus the plain one-word labels of its sibling groups (Fundamentals, Evals, Features, Telephony). Page titles keep the `Pipecat Flows` name people already know, so the sidebar reads **Flows › Overview** while the H1 and unfurl title are unchanged. Same split as Evals, where the group is `Evals` and the page title is `Pipecat Evals`. The Flows overview prose drops the "conversational pathway" phrasing to match.

**The migration guide moves to the Migration group** next to the Pipecat 1.0 guide. The two are a sequence: the Flows guide already tells you to read the Pipecat one first, and the Pipecat one never mentioned Flows, so discovery only worked in one direction. Both pages also carried the sidebar label `Migrating to 1.0`; they're now `Pipecat 1.0` and `Pipecat Flows`.

**That guide was also asserting a release that doesn't exist.** It was titled "Migrating to Pipecat Flows 1.0" and claimed "Pipecat Flows 1.0 requires pipecat-ai >= 1.0.0". There was no Flows 1.0. Per the Pipecat changelog, what actually happened is that Flows moved from the standalone `pipecat-ai-flows` package into `pipecat-ai` 1.5.0 under `pipecat.flows`. The page is reframed around that, and now leads with it. It also picks up two facts it was missing: uninstall `pipecat-ai-flows` (Pipecat logs an error if both are installed), and a pointer to the archived repo where the standalone release history lives.

**`overview/flows.mdx` is dropped** along with its ecosystem card, since Flows is no longer a standalone product. The ecosystem grid is now Pipecat / Clients / Cloud.

**Both "What You Can Build" grids** had a `Conversation Flows` card. In `overview/pipecat.mdx` it linked into the docs while its five neighbours all link to runnable examples; in `pipecat/get-started/introduction.mdx` it pointed at `examples/flows`. Both are now a Multi-Agent Systems card pointing at `examples/multi-worker`.

All 10 old URLs redirect.

## Metadata lint: checks 16 and 17

This PR also changes `scripts/docs-meta-lint.mjs`, because the move could not pass the gate as written.

Check 17 errored on **any** `.mdx` rename or delete, so a legitimate page move could never pass, and check 16's advice to split redirects into their own PR did not help: the renames still trip 17 wherever they land. The two were also mutually unsatisfiable, since 17 wanted a redirect and 16 errored because one had been added.

- **17** now computes the moved page's old URL and requires a redirect from it. The move is fine; the dead inbound link is the error.
- **16** now allows additions and destination rewrites, which a move requires, and errors on the regressions instead: a redirect source disappearing, a new or rewritten entry that resolves to no page, and an untouched entry left dangling by a page this diff moved away. That last case is new coverage the old check could not see, since it only compared the redirect array to itself.

Verified against four perturbations of this branch, each caught:

| Perturbation | Caught by |
|---|---|
| Page moved, its redirect removed | `[17] no redirect from '/pipecat-flows/guides/actions'` |
| Unrelated pre-existing redirect deleted | `[16] redirect removed: '/guides/features/krisp-viva'` |
| New redirect pointing at a nonexistent page | `[16] '/totally-new' -> '/does/not/exist' does not resolve` |
| Page moved, an old redirect left pointing at its former location | `[16] '/guides/features/pipecat-flows' -> '/pipecat-flows/introduction' does not resolve` |

Destination validation is scoped to touched entries and entries aimed at moved pages, rather than all 466, so unrelated PRs are not failed by pre-existing breakage.

Writing that check surfaced one such case: `/examples` pointed at `/pipecat/examples`, which is a directory rather than a page, so the redirect landed on a 404. Fixed here to `/pipecat/examples/overview`, matching its sibling `/recipes` -> `/pipecat/examples/recipes`. **All 466 redirect destinations now resolve to a page.**

## Checks

`mint broken-links` passes. `docs-meta-lint.mjs --git-base main` reports **0 errors**; the 182 warnings are pre-existing.

## Open question

Which `pipecat-ai` release actually removed the deprecated 0.0.x Flows APIs. The changelog documents the 1.5.0 packaging move but not the removals, and that history lives in the archived repo. The page describes *what* to change without asserting a version, which is safe either way, but stating it would be better.


markbackman:
`Conversational Pathways` is a Bland name. We don't want to use that. I think we should just call it out as Flows.

markbackman (approved):
Made a small change, but looks good. This is the right direction.

PR #1131, opened 2026-08-25
Add Gandr published latency numbers to the notes

AALG123:
Follow-up to #1126: the notes now say the first turn is slower without saying by how much. This adds Gandr's published steady-state numbers so the page carries the measured figure instead of nothing:

first audio byte in 146 ms over the open internet, 116 ms p50 first audio, server side warm.

These are the same numbers published on gandr.ai and in gandr.ai/llms.txt, quoted verbatim with their basis.

Disclosure: I maintain the Gandr integration.

### DIFF
--- a/api-reference/server/services/tts/gandr.mdx
+++ b/api-reference/server/services/tts/gandr.mdx
@@ -161,6 +161,8 @@ waiting for that render to finish.
   session voice cache fills. Every turn after it is materially quicker, which
   is why the connection is held warm across turns. A benchmark that measures
   only turn one is measuring the cache filling rather than the engine.
+  Gandr's published steady-state numbers: first audio byte in 146 ms over the
+  open internet, and 116 ms p50 first audio, server side warm.
 </Note>
 
 <Note>


markbackman (approved):

markbackman:
Approving, but this claim assumes a certain geography. I think it's probably highly depend on where the server connecting to your endpoint is located. I would suggest making this clear.

Also, is this TTFB (byte) or TTFA (audio)? TTFA is the first non-silence audio.

Feel free to raise another PR if you need to clarify.

AALG123:
Good questions, answering both:

Byte or audio: these are byte-arrival metrics. That is why the sentence says "first audio byte": the engine's first rendered chunk can be silence, so we do not claim a heard-sound (TTFA) figure.

Geography: agreed on the 146 ms figure, it was measured from a client over the public internet and varies with client location. The 116 ms p50 companion is the same event measured server-side on warm workers, which is location-independent. The two travel together everywhere we publish them precisely so the basis stays attached.

Happy to raise a small follow-up PR making the byte-vs-audio distinction explicit on the page if you want it spelled out there.

PR #1190, opened 2026-08-31
docs: context summarization fires every turn with a large system prompt (T-3682)

jamsea:
## What

A customer had a 24k token system prompt and the default `max_context_tokens=8000`, so context summarization ran on every single bot turn and never helped.

## The gap

Two behaviors are each documented on their own, but the interaction between them is not:

1. The initial system message at `messages[0]` is preserved and excluded from summarization.
2. The auto-trigger token estimate counts every message in the context, system message included.

So when the system prompt alone is near or above `max_context_tokens`, the trigger condition is permanently true. Every bot turn fires a summarization LLM call, and the total can never drop back under the threshold, because the part that is over the threshold is the exact part summarization will not touch.

Verified against `pipecat` main (`src/pipecat/utils/context/llm_context_summarization.py` and `src/pipecat/processors/aggregators/llm_context_summarizer.py`).

## What changed

A short customer-facing warning on the three pages where a reader would look:

- `pipecat/fundamentals/context-summarization.mdx`: warning in **How It Works**, plus one clause in **What Gets Preserved**.
- `pipecat/learn/context-management.mdx`: warning in the **Context Summarization** section.
- `api-reference/server/utilities/context-summarization.mdx`: added to the `max_context_tokens` param description.

Each says the same two fixes: raise `max_context_tokens` above (system prompt + intended history budget), or leave auto summarization off and drive it on demand.

## Note on llms.txt

`llms.txt` and `llms-full.txt` were regenerated per CLAUDE.md. The regen also picks up a `memorysync` page that was already stale on main before this branch, so part of that diff is not mine.

Draft on purpose. Not for merge without review.

### DIFF
--- a/api-reference/server/utilities/context-summarization.mdx
+++ b/api-reference/server/utilities/context-summarization.mdx
@@ -51,7 +51,14 @@ Controls when automatic context summarization triggers.
   Maximum context size in estimated tokens before triggering summarization.
   Tokens are estimated using the heuristic of 1 token per 4 characters. Set to
   `None` to disable token-based triggering. At least one of `max_context_tokens`
-  or `max_unsummarized_messages` must be set.
+  or `max_unsummarized_messages` must be set. The estimate counts every message
+  in the context, including the system message at `messages[0]`, which is
+  preserved and never summarized. Set this value above your system prompt size
+  plus the history you want to keep. If your system prompt alone is close to or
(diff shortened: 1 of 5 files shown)


jamsea:
Hmmm, I feel like we should change pipecat to log a warning message or something in this case 🤔

PR #1227, opened 2026-09-03
docs: add Pipecat Cloud webhooks guide and event reference

Regaddi:
Documents the new Pipecat Cloud webhooks feature. Two pages, added under **Pipecat Cloud > Guides > Webhooks**:

- **`pipecat-cloud/guides/webhooks.mdx`** (Webhooks): registering an endpoint from the dashboard, choosing event types, the signing secret, signature verification in TypeScript (Express) and Python (FastAPI), responding quickly, deduplicating at-least-once retries, the send-test-event action, delivery history, and a troubleshooting table.
- **`pipecat-cloud/guides/webhook-events.mdx`** (Event Reference): the delivery envelope, payload conventions, and all 20 event types with field tables and JSON examples.

No vendor is named except where the wire forces it: the signature header names. Verification is documented against the [Standard Webhooks](https://www.standardwebhooks.com/) specification and the samples depend on `standardwebhooks`, not on the provider's own SDK.

## Signature headers: resolved by capture

An earlier revision of this PR flagged the header prefix as unverified. It has now been settled by capturing a real test delivery through a tunnel:

```
svix-id: msg_3IomK9yQiJnoras33Cc0BzYvoJb
svix-signature: v1,eTiO0rCmMU2tD72X+a6pUgQOTa2ir16AeP7lVlUEZgY=
svix-timestamp: 1788439228
webhook-test: true
```

Deliveries carry the **branded** `svix-` prefix. `standardwebhooks` reads the unbranded names, so both samples now map the three headers across before verifying. The mapping defaults to `""` rather than raising, so an unsigned request fails verification as a 400 rather than erroring as a 500.

Both samples were extracted from the page and executed against a simulated branded delivery: a signed payload verifies, an unsigned one is rejected with a verification error.

## Docs debt this uncovered

`api/docs/webhook-events.md` states the header is `webhook-id`. It is not. That is a third drift in that file, on top of the two the schema module already records against it (`spend.limit_cleared` and the `org.*` payloads). Worth a follow-up in the API repo; this PR documents the wire, not that file.

## Follow-up worth considering, deliberately not done here

The provider supports unbranded headers, which would let the mapping in both samples be deleted. It is a **breaking change** for the subscribers referenced at `api/src/services/webhooks.ts:150`, so it needs a deprecation plan rather than a flag flip. Note also that the delivery `user-agent` names the provider regardless, so this is about ergonomics, not concealment.

## Sourcing

The event reference was written from `api/src/services/webhookEventSchemas.ts`, the machine-readable payload contract that gets published to the delivery provider and that the send-test-event action generates from. Where it disagrees with `api/docs/webhook-events.md`, the schemas win, per the note in that file: `spend.limit_cleared` sends only `previous_limit_cents`.

The setup flow was written from the dashboard's `settings/webhooks` components and `actions/webhooks.ts`. The captured delivery also confirmed the envelope matches what the reference page documents, and that test sends carry `webhook-test: true` while real events do not, which is now noted in the testing section.

## Open question: AWS CodeBuild

The build event descriptions say things like "CodeBuild reports `SUCCEEDED`", naming the build backend. If keeping infrastructure vendors out of the public docs is a general goal, these want rewording too. Left alone as out of scope. Happy to change it.

## Scope notes

- No REST or CLI reference pages. Endpoint management has no private-API-key mount and no `pcc webhooks` command, so it is dashboard-only today.
- `docs-meta-lint` reports the projected `llms.txt` at 100,787 against a 100,000 cap. That was already over before this change (100,325 at baseline) and these pages add roughly 460 chars. A warning rather than a CI error, but the budget is worth a separate look.

## Checks

- `npx prettier --check`: clean
- `node scripts/docs-meta-lint.mjs`: no new errors or findings for these pages (the one remaining error, `memorysync.mdx`, reproduces on a clean tree)
- `node scripts/gen-llms-txt.mjs`: regenerated, committed
- `npx mint broken-links --check-anchors --check-redirects`: no broken links
- Both code samples executed against a simulated branded delivery

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -278,6 +278,13 @@
               "pipecat-cloud/guides/session-api",
               "pipecat-cloud/guides/smart-turn",
               {
+                "group": "Webhooks",
+                "pages": [
+                  "pipecat-cloud/guides/webhooks",
+                  "pipecat-cloud/guides/webhook-events"
+                ]
+              },
+              {
                 "group": "WebSockets",
                 "pages": [
                   "pipecat-cloud/guides/generic-websocket",
@@ -537,7 +544,6 @@
                   },
                   {
                     "group": "Agents",
-
                     "pages": ["api-reference/server/services/agents/openclaw"]
                   },
                   {
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -29298,7 +29298,7 @@ For more details on Smart Turn, see the following links:
 # Pipecat Cloud Webhooks
 Source: https://docs.pipecat.ai/pipecat-cloud/guides/webhooks.md
 
-Register webhook endpoints, subscribe to Pipecat Cloud events, and verify signed deliveries with the Svix SDK in TypeScript and Python.
+Register webhook endpoints, subscribe to Pipecat Cloud events, and verify signed webhook deliveries in TypeScript and Python.
 
 Webhooks let Pipecat Cloud notify your own services when something happens in your organization: a build finishes, a deployment replaces your pods, spend crosses a threshold, a session starts. Instead of polling the REST API, you register an HTTPS endpoint and Pipecat Cloud POSTs a signed JSON payload to it.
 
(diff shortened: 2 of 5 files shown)


markbackman (approved):
Very nice! 👏

PR #1245, opened 2026-09-04
docs: fix Rime PRONOUNCE example and add bulk pronunciation lexicon guidance (T-3868)

jamsea:
Two documentation gaps found while working support ticket 

All claims below were verified against `pipecat` main at `0417f251`.

## Gap A: the Rime `PRONOUNCE` example cannot run

The page showed `RimeTTSService.PRONOUNCE(text, "potato", "potato")`, called on the class with 3 arguments. `PRONOUNCE` is an **instance** method taking `(self, text, word, phoneme)`. The `@staticmethod` decorator above it binds `PAUSE_TAG`, not `PRONOUNCE`. As documented, the call binds `self=text` and leaves `phoneme` missing, so it raises `TypeError`.

`INLINE_SPEED` is an instance method too, and its example on the same page had the identical problem. Fixed both.

Also added:
- A note that `PRONOUNCE` takes three arguments, so it needs a small wrapper function to satisfy the `(text, aggregation_type)` text transform contract.
- A note that `PRONOUNCE` enables bracket phonemization for the next message only, with a pointer to the `Settings` field that keeps it on for the whole session. That is the better choice for anything beyond a one-off.

## Gap B: no guidance on bulk pronunciation lexicons

Nothing in the docs told you how to do a lexicon of hundreds of words, or which providers can host one. New "Bulk pronunciation overrides" section in the Text-to-Speech guide covering:

1. **`replace_text` as the portable client side path.** Rules are regexes compiled once at construction and applied as a loop per chunk, so cost grows linearly and stays small. Measured roughly 0.4 ms per sentence at 300 rules and 1.4 ms at 1000 on an Apple silicon Mac. Presented as an order of magnitude, with hardware variance called out.

2. **The footgun: transforms run after aggregation.** In `TextAggregationMode.TOKEN`, `SimpleTextAggregator` yields each text immediately with no buffering, so a pattern spanning multiple words can never match. Multi-word lexicon entries require sentence aggregation. This is the single most useful thing on the page. Related: pipecat-ai/pipecat#5574.

3. **Provider hosted lexicon table**, checked by grepping every `src/pipecat/services/*/tts.py`:
   - AWS Polly: `lexicon_names`
   - NVIDIA: `custom_dictionary`, with a warning that entries are joined into one comma separated string with no escaping, so a comma in an entry corrupts the dictionary
   - Cartesia: `pronunciation_dict_id`
   - ElevenLabs: `pronunciation_dictionary_locators`, deprecated in 1.6.0 and removed in 2.0.0
   - Azure: not possible. `_construct_ssml` escapes the input text before assembling SSML, so an injected `<phoneme>` tag is escaped and spoken literally, and there is no opt-out. Worth stating, since the page lists "SSML support: Fine-grained pronunciation control" as an advanced feature and people try it on Azure.

4. **ElevenLabs specifics:** prefer respellings over IPA, since respellings work on every model while IPA `<phoneme>` tags need `enable_ssml_parsing=True` and only work on v2 models.

## Checks run

- `npx mint broken-links --check-anchors --check-redirects`: no broken links
- `node scripts/check-imports.mjs --pipecat ../pipecat`: all imports resolve
- `node scripts/docs-meta-lint.mjs`: the one error and the llms.txt budget warning both reproduce on `main` unchanged, so neither comes from this branch
- `llms.txt` and `llms-full.txt` regenerated

Note: `npm run format` is not a no-op on `main` right now, it reformats about 30 unrelated files. I reverted those so this diff stays scoped.

Opened as a draft for review.

### DIFF
--- a/api-reference/server/services/tts/rime.mdx
+++ b/api-reference/server/services/tts/rime.mdx
@@ -348,9 +348,7 @@ tts = RimeTTSService(
 
 ### PRONOUNCE(text: str, word: str, phoneme: str) -> str
 
-Convenience method to support Rime's [custom pronunciations feature](https://docs.rime.ai/api-reference/custom-pronunciation). It takes a word and its desired phoneme representation, returning the text with the provided word replaced by the appropriate phoneme tag.
-
-Unlike `SPELL` and `PAUSE_TAG`, call this method **on the service instance**, not on the class. It also takes three arguments, so it does not match the `(text, aggregation_type)` shape a text transform needs. Wrap it in a small function, the same way the examples above do.
+Instance method that implements Rime's [custom pronunciations feature](https://docs.rime.ai/api-reference/custom-pronunciation). It replaces `word` in `text` with the phoneme tag. Unlike `SPELL` and `PAUSE_TAG`, call it on the service instance, not on the class. Its three arguments do not match the `(text, aggregation_type)` transform shape, so wrap it in a function.
 
 ```python
 tts = RimeTTSService(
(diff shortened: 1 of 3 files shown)


markbackman:
This seems like guide information, not in the learn section, as it's not required for all cases.

PR #1272, opened 2026-09-10
Add acefone serializer

monster-anshu:
The serializer page documented both Acefone and Smartflo (Tata Tele Business). The integration targets Acefone, so drop the Smartflo references and protocol doc link. Regenerate llms-full.txt.


markbackman (approved):
LGTM

PR #1321, opened 2026-09-15
docs: explain why instance count reads higher than max-agents (T-4173)

jamsea:
## Why

A Pipecat Cloud customer deployed with `--min-agents 1 --max-agents 1` and saw `currentRevision.readyReplicas: 2` from the management API with `activeSessionCount: 0`, on every deployment. They read it as a billing and concurrency bug.

It is expected behavior: Pipecat Cloud keeps a small number of spare instances above the limits you configure, as protection against losing an instance (out of memory, eviction, node drain). Nothing public said so.

Worse, two pages implied the opposite, predicting 1 warm agent for this customer at 0 active sessions:

- **Capacity Planning**: "When you start an active agent, Pipecat Cloud automatically provisions an additional idle agent" (so: no extra at zero traffic), plus a warm capacity table reading as an exact count.
- **Scaling > Auto-scaling**: "When your traffic increases and you have active sessions running, our system automatically: Proactively provisions additional idle agent instances".

Neither page told a reader that the instance count they can observe sits above their configured `max-agents` at all times, including with zero traffic.
t hits it.

## Notes

- No internal implementation details are exposed. The multiplier is described qualitatively ("a small number of spare instances") rather than as a number customers could depend on, since it is an internal default we may change.
- Repo-wide grep for the affected claims ("Idle Agents", "warm capacity", "auto-scaling buffer", "readyReplicas", "min-agents", "max-agents", Reserved/Active/Warm tables) found the wrong claim only on these three files. The `pipecat/deployment/**` hits for "warm capacity" are self-hosting pattern pages with no bearing on Pipecat Cloud limits, and the `max-agents` mentions in `error-codes.mdx`, `deploy.mdx` and the CLI reference already describe it correctly as a session cap.


markbackman:
Closing. Making an API change instead.

PR #1328, opened 2026-09-16
Update speechmatics docs to convey the following:

Smiljanic19A:
Update Speechmatics docs to convey the following:

  - Align the endpoint table with the endpoints Speechmatics documents: add the
    global endpoint, keep EU2 as the service default, and drop the neu/wus
    aliases that are not in the public endpoints table. All seven hosts serve
    /v2/agent, so this is about documenting the supported names rather than
    fixing broken ones.
  - Note that global.rt.speechmatics.com routes to the nearest region and may
    route anywhere, so regional endpoints are the choice for data residency, and
    link Speechmatics' supported-endpoints page as the source of truth.
  - Repoint the API reference card at the Agent STT WebSocket reference, which is
    the protocol this service actually speaks.
  - Repoint the Speechmatics documentation card away from the old rt-api-ref path,
    which redirects to the streaming WebSocket reference, to Speechmatics' own
    guide for this integration.
  - Fix the supported-languages link, which only resolved through a redirect.
  - Label operating_point as deprecated since 1.10.0 and note that it resolves to
    model, warns when used, raises ValueError when both are given and differ, and
    is removed in 2.0.0. Mark model as superseding it.
    messages are emitted, not whether partial word fragments appear, and Agent STT
    enables partials when it is unset.
  - Correct the {ts} format attribute from an ISO timestamp to a float of seconds
    from session start. The wire sends metadata.start_time as a number and the SDK
    passes it through without coercion, so an ISO example was misleading.
  - Pin the install command to pipecat-ai[speechmatics]~=1.10.0, the release that
    carries Agent STT support.


api-reference/server/services/stt/speechmatics.mdx
```diff
@@ -149,15 +153,15 @@ Runtime-configurable settings passed via the `settings` constructor argument usi
...
 | `language`               | `Language \| str`            | `Language.EN` | Language code for transcription. _(Inherited from base STT settings.)_                                                                                                    |
 | `domain`                 | `str`                        | `None`        | Domain for Speechmatics API (e.g. for bilingual transcription).                                                                                                           |
 | `turn_detection_mode`    | `TurnDetectionMode`          | `EXTERNAL`    | How turns are closed. `EXTERNAL` has the caller drive turns via `finalize()` (e.g. Pipecat's own VAD); `VAD` lets the STT service run its own VAD and close turns itself. |
 | `speaker_active_format`  | `str`                        | `"{text}"`    | Formatter for speaker output. Available attributes: `{speaker_id}`, `{text}`, `{ts}`, `{lang}`. Example: `"@{speaker_id}: {text}"`.                                       |
```
LArmstrongDev, line 160: This was renamed to speaker_format, if I remember correctly.

markbackman (approved):
LGTM. Let me know once you've ironed out issues and this is ready to go.

api-reference/server/services/stt/speechmatics.mdx
Smiljanic19A: Correct for our other integrations - it was the last thing we did there - by that time PC was merged so we did not change it here

api-reference/server/services/stt/speechmatics.mdx
```diff
@@ -22,11 +22,11 @@ description: "SpeechmaticsSTTService delivers real-time STT with partial and fin
 
 <CardGroup cols={2}>
   <Card
-    title="Speechmatics STT API Reference"
+    title="Speechmatics Agent STT Reference"
     icon="code"
-    href="https://reference-server.pipecat.ai/en/latest/api/pipecat.services.speechmatics.stt.html"
```
LArmstrongDev, line 27: What is this? How is it generated? I checked what's at this link, and some of it is correct, but then it also references things like mulaw etc which are not supported.

api-reference/server/services/stt/speechmatics.mdx
```diff
@@ -149,15 +153,15 @@ Runtime-configurable settings passed via the `settings` constructor argument usi
...
 | `additional_vocab`       | `list[AdditionalVocabEntry]` | `[]`          | Additional vocabulary to boost recognition of specific words.                                                                                                             |
-| `operating_point`        | `Model \| str`               | `None`        | _Deprecated._ Alias for `model`. If both are given they must name the same value, otherwise a `ValueError` is raised. Use `model` instead. Will be removed in 2.0.0.      |
-| `enable_partials`        | `bool`                       | `None`        | Include partial word fragments in partial segment output.                                                                                                                 |
+| `operating_point`        | `Model \| str`               | `None`        | _Deprecated since 1.10.0._ Resolves to `model`, warning when used. If both are given they must name the same value, otherwise a `ValueError` is raised. Removed in 2.0.0. |
```
LArmstrongDev, line 163: What's the 2.0.0 here? That release hasn't been made yet?

api-reference/server/services/stt/speechmatics.mdx
```diff
@@ -149,15 +153,15 @@ Runtime-configurable settings passed via the `settings` constructor argument usi
...
+| `enable_partials`        | `bool`                       | `None`        | Emit interim segments as `AddPartialSegment`, updating until the final segment arrives. Agent STT enables partials when this is unset.                                    |
```
LArmstrongDev:
```suggestion
| `enable_partials`        | `bool`                       | `None`        | Emit interim segments as `AddPartialSegment`, updating until the final segment arrives. Agent STT enables partials by default. Set to false to disable partials.                                    |
```

api-reference/server/services/stt/speechmatics.mdx
```diff
@@ -228,14 +232,14 @@ Examples:
 | ------------ | ---------------------------------- | ------------------------------- |
 | `speaker_id` | The label of the speaker           | `S1`                            |
 | `text`       | The transcribed text               | `Good morning.`                 |
-| `ts`         | The timestamp of the transcription | `2025-09-15T19:47:29.096+00:00` |
+| `ts`         | Start time of the segment, in seconds | `12.34`                      |
```
LArmstrongDev, line 235: Was this a subtle change in the integrations? Going from current timestamp to segment time in audio?

api-reference/server/services/stt/speechmatics.mdx
Smiljanic19A: I do not know - but i do believe our docs are a more fitting reference?

api-reference/server/services/stt/speechmatics.mdx
Smiljanic19A: That was here previously. It was generated in the auto-creation of the docs - which is a part of pipecats process

api-reference/server/services/stt/speechmatics.mdx
Smiljanic19A: The previous service only exposed text and speaker_id, so the ISO example was just wrong.

api-reference/server/services/stt/speechmatics.mdx
LArmstrongDev: True - although previously, it said "will be removed in 2.0.0" - which is in future tense. Now, it's changed to "removed", past tense. That's where it's not clear to me.

api-reference/server/services/stt/speechmatics.mdx
LArmstrongDev: Ah, makes sense. Thanks

api-reference/server/services/stt/speechmatics.mdx
LArmstrongDev: Yes, agreed that the link to our docs is helpful. However, we still need these pipecat docs to be accurate for those users who search in those docs instead.

@markbackman, are you able to share any context on where the reference docs are generated from? Thanks!

markbackman: These are generated to reflect Pipecat's reference API, built by autodoc, here: https://github.com/pipecat-ai/pipecat/tree/main/docs/api

markbackman: Both Pipecat's reference and Speechmatic's reference are relevant here.

api-reference/server/services/stt/speechmatics.mdx
Smiljanic19A: Oh.. did not even see that. That was me - and it was wrong

PR #1329, opened 2026-09-16
docs: say session time filters match on when a session started

Regaddi:
## Summary

- In the Pipecat Cloud REST reference, the agent sessions list's `startTime` and `endTime` now say they filter on when a session started, inclusive at both ends. "After/before this Unix timestamp" didn't say which timestamp of the session they compared.

## Test plan

- [x] `openapi.json` still parses as JSON


mattshep (approved):
## Overview

Reviewed at `f0878c6`.

This rewords `startTime` and `endTime` on the agent sessions list in the REST reference, to say they match on when a session started, inclusive at both ends. It's the docs follow-up to daily-co/pipecat-cloud-sandbox#825.

Approving.

## Verification

**The wording matches the code.** On the API's `main`, the session list filter is now `createdAt >= startTime` and `createdAt <= endTime` for every status. So "started at or after" and "started at or before" are exact, including both inclusive ends.

**This is the only place these parameters are described.** Nothing else in the docs covers the session list's time filters, and the org-wide session list isn't in the public reference.

## Timing (non-blocking)

#825 merged today, but the latest prod API release is still the 09-11 one. Until the next release, prod still applies `endTime` to when a session ended, and not at all to sessions still running, so this wording is ahead of what's live.

Docs only reach production when someone runs the promotion, and that ships everything on `main`. So it's worth merging this when the API release with #825 goes out, rather than before.

## Verdict

Approve. The wording is accurate against the merged code; just time the merge to that API release.

markbackman (approved):

PR #1331, opened 2026-09-16
MoQ transport: reconnect, refused tokens, transcript records, relay_url

mattshep:
Documents the MoQ transport behaviour added in pipecat-ai/pipecat#5800 and pipecat-ai/pipecat-client-web-transports#213.

- **Server page.** Client-mode reconnect (`reconnect_timeout`, the `on_reconnecting` and `on_reconnected` events, what `on_client_disconnected` means once redials are in play), a refused token ending the session without redials, failures reaching the pipeline as an `ErrorFrame` and the role of `processor_unusable_policy`, and `MOQRunnerArguments.relay_url` for hosts that hand the bot a relay URL with a token in it.
- **Transcript records.** Both pages describe the `seq` and `epoch` fields on transcript records and the replay dedupe they enable, written so another client transport on the same stream can implement it.
- **Generated files.** `llms.txt` and `llms-full.txt` regenerated.


vipyne (approved):
lgtm, thx! 👍

PR #1343, opened 2026-09-17
Document optional local Orukeet CPU transcription

Nathan-Roll1:
Documents optional local Orukeet CPU transcription on the existing community-maintained Oruk page. The hosted provider section stays intact. The new section uses an exact public source commit, since the local extra is not in the published release candidate, and explains final-only output, VAD setup, offline caching and the weight license. Companion implementation: https://github.com/Oruk-AI/pipecat-oruk/pull/1.

The adapter downloads pinned model files and required configuration from Hugging Face, verifies SHA-256, and uses normal model-download accounting through `config.json`. Cached inference makes no HTTP requests and sends no audio.

Checks: Mintlify reports no broken links (anchors and redirects enabled), metadata lint has zero errors, and the changed MDX passes Prettier. Generated `llms.txt` and `llms-full.txt` are refreshed. Adapter validation includes 84 passing tests and green Linux Python 3.11–3.14/macOS/Windows checks, plus real English/German/Spanish/French repeats, silence, offline cold reload and two Silero/PipelineWorker utterances on macOS arm64/Pipecat 1.8.1/ONNX Runtime 1.24.4.

The [published NeMo evaluation](https://huggingface.co/oruk/orukeet#evaluation) reports 9.85% versus 11.01% pooled FLEURS WER across 20,146 recordings/25 languages. That source-model result motivates this optional backend; the runtime checks validate the integration separately, without a speed or accuracy comparison. This PR only adds the local-service documentation and matching generated text.


markbackman (approved):
🙌

PR #1345, opened 2026-09-18
docs: update evals for pipecat 1.11.0

aconchillo:
Follow-up to the evals docs updates already on `main` for [pipecat 1.11.0](https://github.com/pipecat-ai/pipecat/blob/main/CHANGELOG.md#1110---2026-09-17). Main already documents the `scenarios:` list, `llm_marker`, `text_excludes`, `eval:` on a `function_call`, `speed`, and the manifest entry options. This PR restructures that coverage so each page explains the basics first and the file-sharing options at the end, and fills in what was still missing.

## Scenario files

- **Scenario Configuration** gets a "Scenario files" section as the single home for the format: the `scenarios:` list, `<file>/<scenario>` naming, file-level defaults with the replace-whole rule, the full list of keys that can sit at the top of a file grouped by kind (`name:` is the one exception), the shared-`turns:` and shared-`persona:` patterns, and a note on the deprecated top-level shape.
- **Scripted Scenarios** and **Simulated Scenarios** each end with a "Several ... in one file" section instead of carrying the multi-scenario examples in their anatomy sections. Each lists that kind's shareable keys with its own examples (`turn_completion`, `interruption`, `book_table`, `diner`).
- The examples on the lifecycle, eval-loop, and quickstart pages are wrapped in the new shape, and sample output shows the `<file>/<scenario>` names and log file names.

## Scripted page

The marker, absence, `text_excludes`, and call-judging material is consolidated under Assertions (main had it split between Events and Assertions, with the call judging in two places). The events table gains `bot_interrupted` and the `vad_*` events, which the marker scenarios anchor on, and `absent:` gets a section of its own.

## Suites and CLI

- An "Entry options" section for `runner_body: {path:|data:}` (with the bare-form deprecation), `name:`, and per-entry `concurrency:`, and a "Concurrency" section for the one-slot-per-entry scheduling and manifest-order dispatch.
- Artifact naming with the entry label and the `<file>__<scenario>` stem, and the `name` and `turns[].expectations` fields in `results.jsonl`.
- CLI reference: `SCENARIOS`, `-p`, `-s`, `-c`, and log and recording naming.

## Library

`EvalScenarioFile.load()` replaces the per-kind loaders throughout, with a "Loading a scenario file" section, `EvalExpectationResult` on turn results, run `name`/`label` in the suite section, the judge's `add_*` and `evaluate_call()` methods, and a "Migrating from 1.10" table for the 1.11.0 deprecations.

`llms.txt` and `llms-full.txt` regenerated. Metadata lint, Prettier, the import check, and `mint broken-links --check-anchors --check-redirects` pass.


markbackman (approved):
LGTM!

